PR #1, opened 2018-02-15
Pacify luacheck; add makefile.

technomancy:
I've found luacheck to be extremely helpful for catching common mistakes.

### DIFF
--- a/.luacheckrc
+++ b/.luacheckrc
@@ -1,0 +1,2 @@
+std = "max"
+-- max_line_length = 80
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,2 @@
+luacheck:
+	luacheck fnl.lua
--- a/fnl.lua
+++ b/fnl.lua
@@ -177,7 +177,7 @@ local function parseSequence(str, dispatch, index, opener)
     local function readValue()
         local start = str:byte(index)
         local stringStartIndex = index
-        local line = currentLine
+        local line = nil
         -- Check if quoted string
         if start == 34 or start == 39 then
             local last, current
@@ -203,7 +203,9 @@ local function parseSequence(str, dispatch, index, opener)
             end
             local forceNumber = rawSubstring:match('^%d')
             if forceNumber then
-                return tonumber(rawSubstring) or error('could not read token "' .. rawSubstring .. '"'), stringStartIndex
+                return tonumber(rawSubstring) or
+                    error('could not read token "' ..
+                              rawSubstring .. '"'), stringStartIndex
             end
             return tonumber(rawSubstring) or sym(rawSubstring), stringStartIndex
         end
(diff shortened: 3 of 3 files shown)


technomancy:
Thanks for the quick merge.

I'm thinking of adding a test suite and improving the documentation for this project. Do you have any plans to continue development of it?

bakpakin:
Not really at the moment, but I would be glad to answer questions or make fixes.

technomancy:
I had a few bug fixes, but I was also thinking of adding in a new function form which could perform arity checks; would you be interested in a patch for that?

bakpakin:
That seems like it would be useful. Do you mean at runtime or at compile time?

technomancy:
Oh heck; as much as I would love to have that kind of data available at compile time as a user, I think it's violently incompatible with the design of the compiler as it currently stands, don't you think?

I see that we already have two function forms, `fn` and `$`. Do you think it would make sense to make it so that one of those does arity checks and one doesn't? My other thought was to do arity checks on `fn` but then introduce `function` which behaves exactly like Lua's. Or we could leave `fn` as-is and introduce `lambda` which throws on arity mismatch. Thoughts?

On an unrelated note, how attached are you to the name "fnl"?

bakpakin:
> Oh heck; as much as I would love to have that kind of data available at compile time as a user, I think it's violently incompatible with the design of the compiler as it currently stands, don't you think?

 I was just checking, but yeah, it is incompatible with the compiler as it stands.

The arity checking as you suggested would be useful. Either syntax is fine, I suppose the former makes the fn keyword more like Clojure.

Just a thought, but if we can do arity checking at runtime, we could also do some basic typechecking
with typehints. Nothing fancy just the basic Lua types to start, but I think that would complement arity checking nicely.

Lastly, I'm not attached to the name at all. This project was originally a stepping stone into another project [dst](https://github.com/bakpakin/dst) that I am currently working on much more actively (which is similarly poorly named.)

technomancy:
> The arity checking as you suggested would be useful. Either syntax is
> fine, I suppose the former makes the fn keyword more like Clojure.

I think it comes down to how comfortable you are introducing runtime
overhead. People are more likely to prefer the shorter forms, so if fn
has arity checks it means the default is to trade speed for safety.

That's the main argument for leaving fn alone and introducing the checks
in a new form like lambda. Personally I could go either way, but it
seems like that's a goal you laid out in the readme.

> Just a thought, but if we can do arity checking at runtime, we could also do some basic typechecking
> with typehints. Nothing fancy just the basic Lua types to start, but I think that would complement arity checking nicely.

That's interesting ... but way more ambitious than I'm ready for right now. =)

However, doing nil-checks might be a good simple solution that gets you
80% of the benefit; what do you think about a naming convention for
parameters that would allow you to indicate that it should be excluded
from the default nil checks?

    (lambda (x y ?a b) ...)

> Lastly, I'm not attached to the name at all.

What do you think about "Deuan" (Thai for "moon" or "month", rendered as
เดือน in the original script)? Another thought was "Fennel" which sounds
similar to the current name and has a pretty flower.

bakpakin:
I have no problem with runtime overhead, as long as it’s optional. Perhaps as you said, one form for native Lua functions, another for more feature rich checked functions. I like the optional tag in the lambda functions.

As for the name, fennel does sound nice. It sounds a bit like I was saying it in my head :).

PR #4, opened 2018-02-16
Added some tests.

technomancy:


technomancy:
There are a few commented-out tests for missing features; I have a half-finished patch which adds `when' and comments. There seems to be a bug when `if' is used for side-effects that might be tricky to track down.

bakpakin:
Alright. Maybe open up a new issue for fixing specific bugs.

PR #5, opened 2018-02-16
Add support for normal semicolon comments.

technomancy:


technomancy:
Thanks.

May I ask what the reasoning is behind the special forms `*for`, `*while`, etc? I noticed `*for` is tied to "numeric" for, and can't work with iterators; is that something that should be added?

Issue #6, opened 2018-02-16
Can't put side-effect inside if branch


technomancy:
```lua
>> (print (fnl.compile "(if true (set x 1))"))
local 
do
  if true then
    x = 1
     = 
  end
end
```

Current workaround is to wrap the branches in a `do`.

bakpakin:
Fixed. Problem was with handling forms that returned no values (the set special form) not promoting
no values to 'nil'

PR #8, opened 2018-02-16
Fix when special form.

technomancy:


technomancy:
Sorry, somehow I missed that there were some test failures on my last PR. This fixes them by adding an additional `nil` to the body of the `do`, but if you have a better solution I'm all ears. (I believe fixing #6 would also fix it, but that one is a bit over my head right now.)

technomancy:
Looks like the fix for #6 took care of this; thanks!

PR #11, opened 2018-02-17
Add lambda for arity-checked functions.

technomancy:
This checks for all non-?-beginning parameters to be non-nil. It doesn't check for cases where there are too many parameters, as I don't believe it's possible to know how many params are passed in unless using var-args.

### DIFF
--- a/fnl.lua
+++ b/fnl.lua
@@ -826,6 +826,22 @@ SPECIALS['$'] = function(ast, scope, parent)
     return SPECIALS.fn({'', sym('$$'), fargs, ast, n = 4}, scope, parent)
 end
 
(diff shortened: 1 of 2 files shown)


bakpakin:
It is actually possible.
```lua
local function returnnothing()
  return
end

local function returnnil()
  return nil
end

print(select('#', returnnothing())) -- prints 0
print(select('#', returnnil())) -- prints 1
```
Shouldn't be hard to add checking for extra parameters, but this is a good start.

technomancy:
That's for checking the number of return values; I was actually thinking about the number of arguments passed in.

So ((lambda [x y] (+ x y)) 1 2 3) should result in an error.

PR #17, opened 2018-02-21
Add tutorial.

technomancy:
I tried to avoid covering anything that's likely to change, with the exception of destructuring multiple values and possibly variadic operators.

Thoughts?


bakpakin:
Wow, looks good!

PR #18, opened 2018-02-21
WIP: Implement method calls with :

technomancy:
A couple things to note here. At first I wasn't even sure we should support method call syntax, but there are enough libraries out there that assume it, I think.

I'm not totally sold on the syntax, but what I came up with was `(: receiver :method ...)` compiles to `receiver:method(...)`. Open for suggestions on how to improve that; perhaps just `(receiver:method ...)` would be better? But that complicates things as it's no longer just another special form.

I couldn't figure out how to indicate to the compiler that the result should be treated as an expression rather than a statement, as you can see from the failing test.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -287,7 +287,7 @@ local function parse(getbyte, state)
             elseif rawstr == 'true' then dispatch(true)
             elseif rawstr == 'false' then dispatch(false)
             elseif rawstr == '...' then dispatch(VARARG)
-            elseif rawstr:match('^:[%w_-]*$') then -- keyword style strings
+            elseif rawstr:match('^:[%w_-]+$') then -- keyword style strings
                 dispatch(rawstr:sub(2))
             else
                 local forceNumber = rawstr:match('^%d')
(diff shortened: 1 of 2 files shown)


bakpakin:
Ideally I would like the `(receiver:method ...)` as well. While that would prevent it from being a special form, the same can be said about the dot syntax.

The `stringMangle` function could be given an extra parameter that is only provided when the symbol is
known to be in calling position. This would allow the last colon character in a symbol to pass unmangled to the compiler. There is similar functionality already for dot syntax, which can me allowed via the multiSym flag to stringMangle.

Keeping track if the symbol was in calling position could be done inside the `compile1` function. Exactly how this would look i'm not sure.

bakpakin:
I have made a few fixes to this get this treated as an expression rather than a statement.

However, this has the same double side-effect problem as variadic comparators. The solution should be much the same.

technomancy:
I think we're agreed this is the wrong approach; closing this and might take a swing at treating it like the dot syntax later.

bakpakin:
Actually I just finished this.

PR #19, opened 2018-02-22
Destructure values, support destructuring in set

### LINKED ISSUE #16: Improve consistency of multi-valued set?
The current way to set multiple values at once is like: `(set a b c (values 1 2 3))`

It seems to me it would be more consistent to support it as `(set a 1 b 2 c 3)` to make it more similar to `let`.

Or another option would be `(set [a b c] [1 2 3])` to make them both support destructuring.

What do you think?

technomancy:
This fixes #16 but it is a breaking change, (for multi-valued `set`) so I'm interested to see what you think about it.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -623,18 +623,28 @@ local function destructure1(left, rightexpr, scope, parent)
     if isSym(left) then
         parent[#parent + 1] = ('local %s = %s'):
             format(stringMangle(left[1], scope), tostring(rightexpr))
-    elseif(isTable(left)) then
+    elseif(isTable(left)) then -- table destructuring
         local s = gensym(scope)
         parent[#parent + 1] = ('local %s = %s'):format(s, tostring(rightexpr))
         for i, v in ipairs(left) do
             local subexpr = expr(('%s[%d]'):format(s, i), 'expression')
             destructure1(v, subexpr, scope, parent)
         end
-    elseif(type(left) == "table" and left.n) then
-        local leftNames = {}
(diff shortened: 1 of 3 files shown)


bakpakin:
I have taken a look at it, and it looks good. I'm not to worried about breaking set, as the semantics weren't that good to begin with.

However, I tried something of the form `(set (a b c) (values 1 2 3))` and it failed. I looked at this and noticed that `destructure` immediately compiled the right form with nval = 1, which means that the returned expressions are forced to have exactly one value. Simply removing this limitation and passing a list of expressions through destructuring instead of a single expression fixed this problem. This problem would not have been noticed with more common code like `(set (a b c) (returns-3-values))`, as a function call is syntactically one expression in Lua.

On another note, I think we should have a local and non-local variant of set. Perhaps `local`, `var`, or something else for the local scope?

technomancy:
Oh cool; nice catch. Thanks for fixing that.

> I think we should have a local and non-local variant of set.

I thought I would really want a `local` form, but after writing a bit of code using `let` for everything doesn't bug me too much. If we do add `local` we might want to consider a form that combines `fn` with it so we can avoid massive indentation as seen in the Pong example. I've always been fond of Scheme's `(define (f a b c) ...)` allowing `define` and `lambda` to be combined, but I don't know if that's a great fit given the way we currently use argument vectors, especially if we want to eventually support destructuring args directly without `let`.

Anyway, I'm not against adding `local`, just noting that its lack doesn't bug me as much as I thought it would.

PR #20, opened 2018-02-22
Allow comparator operators to be variadic.

technomancy:
What do you think of this?

Lack of variadic comparators surprised me a bit. If we don't make them variadic we should at least add arity checks instead of ignoring arguments after the second.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -692,6 +692,18 @@ end
 
 -- SPECIALS --
 
+-- For statements and expressions, put the value in a local to avoid
+-- double-evaluating it.
+local function once(val, ast, scope, parent)
+    if val.type == 'statement' or val.type == 'expression' then
+        local s = gensym(scope)
+        emit(parent, ('local %s = %s'):format(s, tostring(val)), ast)
+        return expr(s, 'sym')
+    else
+        return val
+    end
+end
+
 -- Implements destructuring for forms like let, bindings, etc.
 local function destructure1(left, rightexprs, scope, parent, nonlocal)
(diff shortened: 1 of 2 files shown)


SquidDev:
It's worth noting that this causes the "interior" arguments to be evaluated twice:

```clojure
>> (set debug (fn [x] (print x) x))
>> (< (debug 1) (debug 2) (debug 3))
1
2
2
3
true
```

I'm not sure there's a nice way to do this without creating a closure and temporary variables.

bakpakin:
> It's worth noting that this causes the "interior" arguments to be evaluated twice:

This is actually why I did not implement this originally.

With the new compiler design, however, this is doable without unnecessary allocation of local variables.
The only expressions that would need to be bound to local variables are those with side-effects. The only types of expression with side-effects are the generic 'expression' type and the 'statement' type. The other types ('sym', 'literal', 'varg') can emitted twice without possibility of double side-effect.

The compiler could compile each argument separately to one expression, and if that expression could have side-effects, bind it once to a local variable. The concatenation with and could then be done with the resulting side-effect-free expressions.

This would also make the resulting comparators non-lazy, which is fine, just something to note.

technomancy:
Ah yes, of course.

I'll come back to this later, but for the time being I just pushed a change to ensure comparators always get 2 arguments; better to error out than silently ignoring them.

technomancy:
I think this fixes the double-eval problem described, but I'd appreciate a once-over of the solution before it gets merged. (I factored out the previous double-eval-prevention from the `:` special into a reused `once` function.)

bakpakin:
Looks good to me!

PR #21, opened 2018-02-22
Add partial special form for partial application.

technomancy:
This one might be pushing the limits of minimalism. Partial application is nice to have, but you can easily get it from a library like Lume, so I'm on the fence about whether it should be included or not; totally OK if you decide it's not a good fit.


bakpakin:
Yes, lume may do it, but its useful and very little overhead.

PR #23, opened 2018-02-25
Line numbers for compile errors

technomancy:
This doesn't solve "line numbers" in general, but for compile-time errors we should have enough data to emit error messages with useful filename/line number details.

This uses `line` and `filename` as top-level locals and adds a `line` field to all AST lists created in the parser. The ast is passed to `assertCompile` which uses it to construct a message that's a bit more helpful. The `assertCompile` function also ensures that the error message includes the name of the form being compiled.

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,12 +58,19 @@ chunk when called. Useful for implementing a repl.
 local bytestream, clearstream = fennel.granulate(chunks)
 ```
     
-Get the next top level value parsed from a stream of
-bytes. Returns true in the first return value if a value was read, and
+Converts a stream of bytes to a stream of values.
+Valuestream gets the next top level value parsed.
+Returns true in the first return value if a value was read, and
 returns nil if and end of file was reached without error. Will error
 on bad input or unexpected end of source.
 ```lua
-local ok, value = fennel.parse(strm)
+local valuestream = fennel.parser(strm)
+local ok, value = valuestream()
+
+-- Or use in a for loop
+for ok, value in valuestream do
+    print(ok, value)
+end
 ```
 
 Compile a data structure (AST) into Lua source code. The code can be loaded
--- a/fennel
+++ b/fennel
@@ -17,14 +17,18 @@ if arg[1] == "--repl" or #arg == 0 then
 elseif arg[1] == "--compile" then
     for i = 2, #arg do
         local f = assert(io.open(arg[i], "rb"))
-        print(fennel.compileString(f:read("*all")))
+        local ok, val = pcall(fennel.compileString, f:read("*all"))
+        print(val)
+        if(not ok) then os.exit(1) end
         f:close()
     end
 elseif #arg == 1 then
     local filename = table.remove(arg, 1) -- let the script have remaining args
     local f = io.open(filename, "rb")
     if f then
-        fennel.eval(f:read("*all"), {filename = filename})
+        local ok, val = pcall(fennel.eval, f:read("*all"), {filename=filename})
+        print(val)
+        if(not ok) then os.exit(1) end
         f:close()
     else
         print(help)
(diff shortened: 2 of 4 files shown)


bakpakin:
Looks pretty good for the most part.

My one concern right now is the keeping of the current line in essentially a global variable. Perhaps we could encapsulate that in the state that is passed to the parse function.

bakpakin:
Actually, I think a better option would be to wrap parse in a closure, allowing easy capture of state and better use as an iterator. Something like:

```lua
local function parser(getbyte)
  local line = 1
  return function ()
    -- Parse code here
  end
end
```

technomancy:
Moving `line` to a local is pretty straightforward, (the `state` parameter for `parse` seems to be redundant and should probably be removed). I don't see why making the parser function close over `line` would be helpful over just making it a regular internal local; maybe I'm missing something?

But moving filename data to a local is less so since it needs to be passed all the way from `compile` to `assertCompile`. Are you OK with leaving it as an effectively-global, or would you rather try to add it to each AST node?

bakpakin:
I am currently testing changing the `parse` function to the `parser` function. It seems rather straight forward and fits rather nicely with the stream of bytes idea. Parser would take a stream of bytes and return a stream of values.

technomancy:
> The current line needs to be persisted over multiple calls to parse, so it needs to be preserved somehow.

I'm not sure I follow this; `parse` is only called from `compileStream`. Would `compileStream` be called multiple times on the same stream without resetting lines from the beginning? Or are you just planning ahead for future changes?

technomancy:
I'll take a shot at moving filename data to the AST tables.

bakpakin:
Chnages would be minimal internally. Instead of calling `parse(getbytes)` directly could do:

```lua
local read = parser(bytestream)
local ok, val = read()
```

technomancy:
Yeah, good call; that does end up looking a lot nicer in the end.

I'm also thinking (and I'm sure you've had the same thought) about runtime line numbers. I am somewhat skeptical that it will work, but I'm thinking about adding a mode where the compiler attempts to only include newlines in the output when they're present in the input. Obviously it won't be perfect given that macros can change the ordering of forms, but I think I might give it a shot in another branch.

bakpakin:
That is an interesting idea. It would make the generated code messy, but that should be solvable.

My original thought was to append comments to mapped lines that contained the line information, and a custom error handler could somehow extract that. I would also prefer to evetualy keep byte offsets as well, and keep track of both the start and end offsets for forms.

As for the macro problem, I'm pretty sure there is no simple perfect simple solution as primitive values are 'unboxed' in Lua (numbers, booleans, and strings), but for all other types we can attach the source metadata somehow (symbols and table types) in the parser, as we do now for lists and tables (basically do this for symbols as well). I'm still unsure of the best way to handle primitives.

Lastly, where we do:
```lua
parent[#parent + 1] = 'Lua code here...'
```
we would need to replace with 
```lua
emit(parent, 'Lua code here', sourceMeta)
```
where sourceMeta could contain line information for the most specific form that generated this section of code. The emit function could then append the formatted comments to the source, or somehow store it
for later appending.

technomancy:
I actually have already written `emit` just as you described, but it always takes `ast` instead of `sourceMeta`. This would at least allow each open paren to line up, so maybe I'll give that a shot first and then maybe see if we need to plumb thru the symbol metadata as well later.

There are a couple places where I can't do that `emit` replacement, when a special form returns a value instead of splicing things into `parent`; for instance in `SPECIALS['.']` and in the `iife` section of `SPECIALS['if`]`. Can you explain why these are different?

bakpakin:
These simply return expressions, which are eventually spliced in by the calling code, usually `compile1`, `handleCompileOpts`, and `keepSideEffects`. If these functions fully use emit, I see no reason why all specials shouldn't work.

technomancy:
That makes sense; I hadn't gotten to those other functions yet.

PR #24, opened 2018-02-26
Attempt to correlate line numbers in output with inputs

technomancy:
A rough first pass; needs a lot more testing but appears to not blow things up too badly.

Currently only active when you pass `{accurate=true}` as options to `compile` or `eval`.

Interested in hearing what you think. It turned out easier than I expected.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ test:
 	$(LUA) test.lua
 
 luacheck:
-	luacheck fennel.lua
+	luacheck fennel.lua fennel
 
 count:
 	cloc fennel.lua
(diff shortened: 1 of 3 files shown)


technomancy:
Actually as I think about it more, if the purpose of this feature is to improve the stack traces, then I don't see how preserving line data for symbols is important, because stack trace frames can only come from calls to functions or special forms, and those are always lists.

Am I missing some other case where line number info from symbols (or even strings or numbers) could be relevant?

bakpakin:
I haven’t played around with this pull request yet, but that is an interesting point. I suppose it is not necessary as of now, but it seems to me necessary if we want to solve reordering from macros, in which case we can’t rely on a 1 to 1 mapping.

PR #29, opened 2018-03-05
Require

technomancy:
I've added a few features on this branch. First of all I think `fennel.dofile` is pretty nice to have; it just moves some stuff out of the wrapper executable so others can re-use it.

I started to add a `fennel.require` script, but then I looked into how Lua implements regular `require` and it's designed to be extensible by adding new entries into the `package.loaders` table. So I gave that a shot instead. This patch implements a `fennel.searcher` function which you can put there to make regular `require` work transparently on Fennel code:

```lisp
>> (set fennel (require "fennel"))
>> (table.insert package.loaders fennel.searcher)
>> (require "hey.in") ; runs (fennel.dofile "hey/in.fnl") on the first run and caches for subsequent runs
188.4
table: 0x563bf92663f0
```

This seems pretty nice, but I'm not 100% sure we want Fennel modules to be able to shadow Lua modules? I guess there are pros and cons. We could also just implement our own distinct `fennel.require` function which doesn't touch the built-in Lua `package` table at all and implemented its own version of `package.loaded`. I'm happy to implement that if it's what you prefer.

Thoughts?

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -1323,6 +1323,12 @@ local function eval(str, options)
     return loader()
 end
 
+local function dofile_fennel(filename)
+    local f = assert(io.open(filename, "rb"))
+    eval(f:read("*all"), { filename = filename, accurate = true })
+    f:close()
+end
+
 -- Implements a configurable repl
 local function repl(givenOptions)
     local options = {
@@ -1398,7 +1404,8 @@ local module = {
     scope = makeScope,
     gensym = gensym,
     eval = eval,
-    repl = repl
+    repl = repl,
+    dofile = dofile_fennel,
 }
 
 SPECIALS['eval-compiler'] = function(ast, scope, parent)
--- a/fennel.lua
+++ b/fennel.lua
@@ -1336,7 +1336,6 @@ local function repl(givenOptions)
         read = io.read,
         write = io.write,
         flush = io.flush,
-        print = print,
     }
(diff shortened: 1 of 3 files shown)


technomancy:
To be clear, I'm totally fine leaving out `require` support altogether, though I do think `dofile` would be nice.

bakpakin:
I definitely think this is a good idea and will merge it, both the require and the dofile.

I see no problem with shadowing Lua files, can you elaborate?

technomancy:
Oh, it's just that I don't know the specific details of how Lua defines modules; if there's supposed to be a 1:1 correspondence with the argument to `require` and a chunk of Lua code then it could be ill-advised, but I suppose less of one if the application is responsible for installing our searcher into `package.loaders`.

But I asked in `#lua` on freenode and they seem to agree it's not a problem, so yeah, probably worrying over nothing.

fennel.lua
```diff
@@ -1323,14 +1323,21 @@ local function eval(str, options)
     return loader()
 end
 
+local function dofile_fennel(filename)
+    local f = assert(io.open(filename, "rb"))
+    local vals = tpack(eval(f:read("*all"), { filename = filename,
```
daurnimator: Do the read; then close; then do the eval as a tail call. Should be much faster :)

technomancy: Good call; thanks.

PR #30, opened 2018-03-06
Fix (some) macros

technomancy:
I found a few bugs in the `macro` implementation which this PR fixes, but it doesn't go all the way towards fixing it. There's still a problem where `COMPILER_SCOPE` is wrapped in a subscope, so `emit(parent, ('_SCOPE.specials[\'%s\'] = %s'):format(macroName, s), ast)` sets a `specials` which is just discarded when the form is done.

I'm not sure how to ensure that the top-level specials table is the one that gets the new entry, but this should at least get it closer to working.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -965,15 +965,15 @@ end
 SPECIALS['macro'] = function(ast, scope, parent, opts)
     assertCompile(scopeInside(COMPILER_SCOPE, scope),
                   "can only declare macros in 'eval-compiler'", ast)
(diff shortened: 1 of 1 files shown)


technomancy:
There are still problems with mangling of names (defining a macro named `->` doesn't work) but I think this latest commit fixes it for names that don't need mangling.

technomancy:
Thanks! I think I have an understanding of how to fix the mangling too now; just needs a bit more testing and I'll make another PR.

BTW I've started a `#fennel` channel on Freenode, and there are a few of us chatting there if you want to join.

PR #31, opened 2018-03-07
Add integer division operator (//) to arithmetic specials

Benaiah:


bakpakin:
I would prefer to have some kind of fallback for Lua versions less than 5.3, but I will merge this for now. I am slow to respond as I am on vacation.

PR #32, opened 2018-03-08
Fix special forms, allow mangled macro names to work.

technomancy:
Add tests for macros and special forms.

This adds `list` and `sym` to the `eval-compiler` env, breaks the current convention of only introducing `_THIS_STYLE` entries, but I think it's necessary, otherwise most macros are basically impossible to write.

Want to make sure you're OK with introducing these new globals into the macro scope before merging this.


technomancy:
Also, let me know if you would like a continuous integration setup; I work for CircleCI and could set it up so all the tests run on every commit and PR if that's something you want to happen.

bakpakin:
Yeah, a circle CI pipeline would be nice. 

I will merge this pull request and don’t worry about the naming convention. My original intent was to expose the entire fennel module via _FNL, but I suppose that could get tedious for basic useful functions.

PR #36, opened 2018-03-13
Treat nil as a self-evaluating symbol in the AST instead of actual nil.

### LINKED ISSUE #35: incorrect output on let binding
this fennel file:
```
(let [x 1
      r nil
      l 3]
      (print ""))
```
generates:
```
do
  local x = 1
  local r = nil
  return print((""))
end
```
this does not generate a local l var, leaking any later assignments to the global namespace.

with some testing i noticed that all let assignments after a nil assignment are ignored. So in the above example x is emitted but l is not.

Changing the assigment to `r false` fixes the problem

technomancy:
Given that Lua tables cannot store nil and lisp code is stored in data
structures, we either must cheat on tables and try to find a way to
make them store nil, or cheat on nil and find a way to store it in tables.

I believe the second approach is simpler since the symbol `nil`
naturally evaluates to nil; the only hiccup is that it must be
disallowed from being used as the name of a local. There are only two
places function parameters are introduced (destructuring and fn) so
catching those two is pretty straightforward.

I believe this would allow us to clean up a lot of AST-using code
since we could now trust Lua's built-in length operator to do the
right thing and stop tracking the .n field of parsed lists. Having to
ensure the correct value in the .n field was one surprising thing I
found quite error-prone when writing my own macros.

Fixes #35.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -298,7 +298,7 @@ local function parser(getbyte, filename)
                 until not b or not issymbolchar(b)
                 if b then ungetb(b) end
                 local rawstr = string.char(unpack(chars))
-                if rawstr == 'nil' then dispatch(nil)
+                if rawstr == 'nil' then dispatch(sym("nil"))
                 elseif rawstr == 'true' then dispatch(true)
                 elseif rawstr == 'false' then dispatch(false)
                 elseif rawstr == '...' then dispatch(VARARG)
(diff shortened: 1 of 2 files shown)


technomancy:
The second commit here isn't strictly needed, but I think it cleans up the implementation nicely. Oddly enough it passes the tests on every lua except LuaJIT, so I'll look into why that's happening.

technomancy:
Neat; I had forgotten that `expr` was a thing, but yeah that makes more sense here. Thanks.

bakpakin:
This does work better than I would have thought and certainly strives a bit better towards minimalism than my original solution.

PR #37, opened 2018-03-15
Replacement require-macros form

technomancy:
After some discussion on the `#fennel` freenode channel it was
observed that the current `macro` and `special` calls could be
problematic on larger codebases because they are global.

I started thinking about what a macro system would look like which
used the existing module system and didn't affect global compiler
state, and I came up with the `require-macros` form.

```lisp
;; main.fnl
(require-macros "mymacros")

(defn add-seven-then-multiply-by-a-hundred [x]
  (-> x (+ 7) (* 100)))

(print add-seven-then-multiply-by-a-hundred 93)
```

```lisp
;; mymacros.fnl
{"->" (fn [val ...]
        (each [_ elt (pairs [...])]
          (table.insert elt 2 val)
          (set elt.n (+ 1 elt.n))
          (set val elt))
        val)
 "defn" (fn [name args ...]
          (list (sym "set") name
                (list (sym "fn") args ...)))}
```

This allows you to import macros just for the scope of a single file
(or if you really want to, for the scope of a single function). The
`mymacros.fnl` file would be loaded in the same kind context as
`eval-compiler` currently works.

I *think* it's comprehensive enough that it could replace the existing
`macro`, `special`, and `eval-compiler` forms, but I'll admit I
haven't thought thru the implications of that fully. Maybe `special`
and `eval-compiler` are still needed in some contexts?

What do you think?

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,8 +54,10 @@ Takes these additional options:
 ### Evaluate a file of Fennel
 
 ```lua
-local result = fennel.dofile(filename)
+local result = fennel.dofile(filename[, options])
 ```
+
+* `env`: same as above.
 
 ### Use Lua's built-in require function
 
@@ -151,7 +153,7 @@ local lua = fennel.compile(ast)
 
 ## Try it
 
-Clone the repository, and run `lua fennel --repl` to quickly start a repl.
+Clone the repository, and run `./fennel --repl` to quickly start a repl.
 
 ## Install with Luarocks
 
@@ -176,6 +178,7 @@ fennel --compile myscript.fnl > myscript.lua
 ## Resources
 
 * [Emacs support](https://gitlab.com/technomancy/fennel-mode)
+* [Wiki](https://github.com/bakpakin/Fennel/wiki)
 * Build: [![CircleCI](https://circleci.com/gh/bakpakin/Fennel.svg?style=svg)](https://circleci.com/gh/bakpakin/Fennel)
 
 ## License
--- a/fennel.lua
+++ b/fennel.lua
@@ -1321,11 +1321,13 @@ local function eval(str, options)
     return loader()
 end
 
-local function dofile_fennel(filename)
+local function dofile_fennel(filename, options)
+    options = options or { accurate = true }
     local f = assert(io.open(filename, "rb"))
     local source = f:read("*all")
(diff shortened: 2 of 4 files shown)


bakpakin:
I think this a good start. I have never liked the clojurescript idea of require-macros, but it does keep macros modular. I would eventually like require-macros to look more like normal fennel.

I think that the macro/extension system is overly complicated, but I think I would like to keep eval-compiler no matter what. While not often needed, it can be used to extend the compiler in pretty much arbitrary ways.

Perhaps the `special` special form could be dropped though.

technomancy:
Thanks.

> I would eventually like require-macros to look more like normal fennel.

Can you go into more detail here? As far as I'm concerned the modules
loaded by `require-macros` are normal Fennel; the only difference is a
few extra entries in the environment, which is the same as how
`eval-compiler` works.

Do you mean you'd like to allow them to exist in the same file instead
of requiring module isolation?

> I think that the macro/extension system is overly complicated, but I think I would like to keep eval-compiler no matter what.

That makes sense.

Originally I was thinking there would be no need for new specials
outside what ships with Fennel, because the only things that couldn't be
done as a macro would be things that emitted Lua keywords, and there are
only a known fixed set of those. But new additions could be added to
Lua, or implementation-specific extensions that might only make sense to
emit in, say, erlua or luaj. Or crazy things like "for the purposes of
this repl session, replace `local` with `set`".

So keeping `eval-compiler` makes sense. But yeah, `special` is
unnecessary when you consider that `_SPECIALS` is in scope inside
`eval-compiler`.

bakpakin:
> Can you go into more detail here? As far as I'm concerned the modules
loaded by `require-macros` are normal Fennel; the only difference is a
few extra entries in the environment, which is the same as how
`eval-compiler` works.

I was thinking more along the lines of looking like common lisp, with a defmacro kind of syntax, but with lexical scoping. I guess originally I had envisioned macros and functions living along side one another in the same file, but that was never set in stone.

Something like:
```lisp
(do
  (defmacro mymac [x] (list (sym "if") x x))
  (mymac :hi))
```

Which would expand to

```lisp
(do
  (if :hi :hi))
```
Without modifying the environment.

PR #40, opened 2018-03-18
Add fennelview for pretty-printing repl results.

technomancy:
This change uses `pcall` to attempt to load `fennelview.fnl` as the default pretty-printer for the repl, but it safely falls back to `tostring` if it's not found. This allows `fennel.lua` to continue to be used as a single-file library.

I added generative testing, but I'm not sure it's quite ready to merge. Certain seeds will produce strings which crash the parser (`SEED=1521395122`, `SEED=1521395046`) and other seeds can cause infinite loops in `fennelview`. (`SEED=1521395257`)

So I'll try to get those sorted out before merging.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,14 @@ fennel.repl([options])
 Takes these additional options:
 
 * `read`, `write`, and `flush`: replacements for equivalents from `io` table.
-* `pp`: a pretty-printer function to apply on values; defaults to `tostring`.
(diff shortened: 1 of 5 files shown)


bakpakin:
For the seeds that crash the parser, we can try to fix the parser if you create an issue

technomancy:
Amended to fix a couple bugs; once #41 is fixed this will be ready to merge.

PR #42, opened 2018-03-19
Support destructuring the rest of a table into a local.

technomancy:
How does this look? I just copied the notation from Clojure; I'm not stuck on it if there's something else you'd prefer.


bakpakin:
Awesome!

PR #43, opened 2018-03-20
Opt-in local mutability and separate global setter

technomancy:
I'm a little hesitant about this change because it's not backwards-compatible, but one of the most common mistakes with Lua is accidentally making a typo in an assignment and creating a global when you meant to just give a new value to a local. That's a big part of why `luacheck` is such an important tool.

But this is only a problem because Lua uses the same syntax for setting globals as for setting locals. What if we limited `set` so that it could only give a new value to a local that had already been introduced and used `set!` to change the value of globals?

I've implemented that behavior in this pull request. It's a big change, so it's worth thinking thru. For folks who have existing code, it should be pretty easy to update as the error message is pretty descriptive: `Compile error in `set' in.fnl:3: tried to set local row_str that wasn't in scope; use set! for globals`

But maybe it would be better to leave `set` alone and introduce a new form which only works on locals, and encourage using that where possible? I can't think of a better name though.

### DIFF
--- a/README.md
+++ b/README.md
@@ -151,10 +151,10 @@ local lua = fennel.compile(ast)
 
 #### Fibonacci sequence
 ```
-(set fib (fn [n] (or (and (> n 1)
-                          (+ (fib (- n 1))
-                             (fib (- n 2))))
-                     1)))
+(local fib (fn [n] (or (and (> n 1)
+                            (+ (fib (- n 1))
+                               (fib (- n 2))))
+                       1)))
 
 (print (fib 10))
 ```
@@ -184,6 +184,8 @@ To compile a file:
 ```sh
 fennel --compile myscript.fnl > myscript.lua
 ```
+
+When given a file without a flag, it will simply load and run the file.
 
 ## Resources
 
--- a/fennel.lua
+++ b/fennel.lua
@@ -1015,12 +1015,16 @@ SPECIALS['set!'] = function(ast, scope, parent)
     destructure(ast[2], ast[3], scope, parent, true)
 end
 
+local function inScope(name, scope)
+    return scope.manglings[name] or scope.parent and inScope(name, scope.parent)
+end
+
 SPECIALS['set'] = function(ast, scope, parent)
     assertCompile(#ast == 3, "expected name and value", ast)
     local target = ast[2][1]
     local parts = isMultiSym(target)
     if parts then target = parts[1] end
-    assertCompile(scope.manglings[target],
+    assertCompile(inScope(target, scope),
                   ("tried to set %s which isn't in scope; use set! for globals")
                       :format(target), ast)
     destructure(ast[2], ast[3], scope, parent, true)
--- a/fennel.lua
+++ b/fennel.lua
@@ -1010,8 +1010,19 @@ SPECIALS['.'] = function(ast, scope, parent)
     return ('%s[%s]'):format(tostring(lhs[1]), tostring(rhs[1]))
 end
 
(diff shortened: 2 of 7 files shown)


bakpakin:
I do agree that set should only work on values that are in a lexical scope. Fennel is too young to be too much concerned with backwards compatibility, so we should definitely add it.

> But maybe it would be better to leave set alone and introduce a new form which only works on locals, and encourage using that where possible? I can't think of a better name though.

Perhaps `def` could be used instead of `set!` to indicate that this is in a global scope. Or maybe the keyword `global(!)`.

I have also been thinking about having to kinds of locals - constants and mutable locals (defs and vars). The only difference would be that the set command would only work for vars and error if the programmer tries to set a def. I'm interested in what you would think of this.

bb010g:
I like the idea of having to declare globals too, as it mirrors `local` usage. So, for the [pong](https://p.hagelb.org/pong.fnl.html) example, you're looking at just a `(global love)` at the top. At that point, you could enforce `set!` to only work on globals and always be explicit.

`def` for local constants sounds good (global constants don't make much sense, as you can't convey constantness across `require`).

technomancy:
> Perhaps def could be used instead of set! to indicate that this is in
> a global scope. Or maybe the keyword global(!).

I would expect `def` to indicate the introduction of a new global rather
than giving an existing global a new value, but such a distinction isn't
really possible to deliver on. But IMO `global` is even better as it
stands out more as a "you probably don't want this" thing.

> I have also been thinking about having to kinds of locals - constants
> and mutable locals (defs and vars). The only difference would be that
> the set command would only work for vars and error if the programmer
> tries to set a def. I'm interested in what you would think of this.

I like the sound of this! I keep thinking that Fennel must work like
Lua, but we should take advantage of the fact that we do have enough
context tracked by the compiler to spot common errors and make them
impossible. Opting into local mutability is a great correctness help.
At first I was thinking "oh, eventually I'll want to port luacheck to
Fennel" but now I'm thinking more along the lines of "how can we make
luacheck unnecessary?"

I'm imagining a new `var` special which would work just like `local`
except it would also set a flag in the `scope` which would act as a
whitelist for `set`; is that roughly what you had in mind? I'd like to
make `set` opt-in such that function arguments and `let`-bound locals
also wouldn't work with it.

bakpakin:
Yeah, this is pretty much what I had in mind.

bb010g:
> I'm imagining a new `var` special which would work just like `local` except it would also set a flag in the `scope` which would act as a whitelist for `set`… I'd like to make `set` opt-in such that function arguments and `let`-bound locals also wouldn't work with it.

So where does this leave `local`? Would `local` be renamed to `var`?

technomancy:
> So where does this leave local? Would local be renamed to var?

From a coding perspective, var would work the way local does now.

From an implementation perspective, var would be local plus "flag as mutable".

technomancy:
I've pushed out the changes suggested; thoughts on the implementation?

This makes it so globals can only be set with `global`, and `set` only works on locals that have been introduced with `var`. I haven't updated the docs yet but will do that once we agree on the way it should look.

bakpakin:
Yeah it looks great, implementation is nice and light!

technomancy:
Cool; thanks. I'll update docs tomorrow and merge then.

It's amusing to think that this same feature was proposed for Java 10 but they weren't able to agree on how it should work in time for the release.

technomancy:
Got some docs in place for this and feeling pretty good about it.

PR #46, opened 2018-03-28
Add a whole bunch of documentation.

technomancy:
* Move API docs to a separate file
* Add reference listing out all special forms
* Add Lua primer with an overview of Lua concepts and pointers to the manual

I still want to improve the examples in the readme.

Do you think it's worth keeping the Luarocks section around? I suspect most people would be better off using git instead since there are no dependencies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,11 +4,6 @@ Fennel (formerly fnl) is a lisp that compiles to Lua. It aims to be easy to use,
 zero overhead compared to handwritten Lua. It's currently a single file Lua library that can
 be dragged into any Lua project.
 
-See [the tutorial](https://github.com/bakpakin/Fennel/tree/master/tutorial.md)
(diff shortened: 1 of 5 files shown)


technomancy:
Yeah, I left a few things undocumented.

* `eval-compiler`: this one is just hard to explain, so I figure we can come back to it later along with `require-macros`
* `special` and `macro`: I think these will probably be removed eventually once we settle on something we like
* `*while`: should probably be renamed to `while`?
* `$`: I'm not actually sure how this works; shorthand for fn? but I've never felt like fn is too long.
* `pack`: seems redundant given `[x y z]` does the same thing?
* `block`: seems redundant with `do`?
* `luastatement`, `luaexpr`: do these just let you include Lua code literally inline? Not sure how they're intended to be used; maybe as a kind of escape hatch which is simpler than `eval-compiler`?

PR #50, opened 2018-03-30
Make --accurate the default option

technomancy:
I think it makes sense to turn this on by default for all entry points that actually run code.

Here I have it so you can pass `--indent "  "` to the `--compile` command... It might make more sense to make the nicely-indented output the default for `--compile` and not for the others, but I decided to err on the side of consistency. Hopefully as the compiler matures and the bugs get shaken out, the need for readable Lua output will be reduced.

### DIFF
--- a/api.md
+++ b/api.md
@@ -1,9 +1,6 @@
 ## Fennel's Lua API
 
-The fennel module provides the following functions. Most functions take
-an `indent` function to override how to indent the compiled Lua output
-or an `accurate` function to ignore indentation and attempt to make
-the lines in the output code match up with the lines in the input code.
+The fennel module provides the following functions.
 
 ### Start a configurable repl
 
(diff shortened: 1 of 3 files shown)


bakpakin:
I will merge this for now, but I would rather have readable compiled output than not. I'm also unsure if the `--accurate` flag for source mapping is robust and wonder if it will handle macros well. I think long term, embedding comments in the generated source with mapping information is the more robust way to go.

technomancy:
Sure; I think it's reasonable to default to `--indent "  "` when using the `--compile` command. I was kind of on the fence about that part. I just didn't think it was good for the repl and "run this file" commands, because in those cases the output was never actually visible anywhere.

I can go change the behavior back for compile. Do you think it should be set in the `compile*` functions of `fennel.lua` or in the `fennel` wrapper?

bakpakin:
Probably the compile functions. The indentation really doesn't hurt and it is nice to have when doing compiler debugging.

technomancy:
Good call; pushed a fix. `eval` now preserves lines and `compile*` functions don't.

PR #51, opened 2018-03-30
Remove some unused specials and functions.

bakpakin:
Remove block, $, pack, and rename *while to while.
Expose fennel module at compile time.

I have removed some of the unnecessary special forms and also made it clear that in the eval-compiler
context, the fennel module is exposed. This means that in require-macro, you can do stuff like `(fennel.eval "(print :hi)")` for your extra meta needs.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -26,7 +26,6 @@ local pairs = pairs
 local ipairs = ipairs
 local tostring = tostring
 local unpack = unpack or table.unpack
-local tpack = table.pack or function(...) return {...} end
 
 --
 -- Main Types and support functions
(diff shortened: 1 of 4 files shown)


bakpakin:
Also, the `:` special somehow was in the code twice, so I removed the first unused definition.

bakpakin:
> Out of curiosity, is when you say while is "preferable to a lambda function and tail recursion", is that just because it avoids function call overhead? I saw a claim somewhere that TCO can in some cases be faster than loops, but I haven't tested it myself.

Yeah. It can also prevent creation of a closure, and often it is the clearer thing to do.

PR #52, opened 2018-04-03
Sort non-integer keys in tables

Benaiah:
The current system uses `pairs` to iterate through non-numeric keys, which causes successive compilations of the same file to differ from one another, since `pairs` is an interface that explicitly does not specify ordering. The specific use case that prompted this PR was trying to do fennel compilation in a git pre-commit hook - since compilation is currently non-idempotent, you can't repeatedly compile the same file to get the same result.

This PR simply compiles key values in the AST to strings, sorts them using Lua's default string sorting, and then iterates through the sorted key list to print the keys and values. All current tests pass, and the compiler is now idempotent for every file I've run it on.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -678,12 +678,17 @@ local function compile1(ast, scope, parent, opts)
         for i = 1, #ast do -- Write numeric keyed values.
             buffer[#buffer + 1] = tostring(compile1(ast[i], scope, parent, {nval = 1})[1])
         end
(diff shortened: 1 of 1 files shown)


bakpakin:
Thanks a bunch. Note that Fennel will likely never have completely deterministic compilation due to table literals, but this is definitely a step in the right direction where the common cases are deterministic.

PR #57, opened 2018-04-04
A version of the thread last macro, AKA ->>.

thatismatt:
This is the version of `->>` that I've been using locally, and it appears to work(!). But as I just copied from the implementation of `->` I'm not sure that it is correct. In particular I wasn't sure about this line:

    (set elt.n (+ 1 elt.n))


bakpakin:
That should be fine. That is a hold over from when the parser would could add embedded nils to the source code, so we used .n to track the length of list like tables. Since we insert into the form, we increase the length.

PR #58, opened 2018-04-04
defn - use local or set (not global)

thatismatt:


technomancy:
Yes; definitely better than making it global. But what if it could behave differently when given a dotted symbol, such that `(defn love.update [dt] ...)` would use `set` instead of `local`?

thatismatt:
That'd be nice, and would avoid this "define & export" dance:
```
(defn a ...)
(defn b ...)
(defn c ...)
{:a a :b b :c c}
```

bb010g:
How's this look?
```
  :defn (fn [name args ...]
    (local form (if (multi-sym? (. name 1))
                    (sym :set)
                    (sym? name)
                    (sym :local)
                    ; else
                    (error "defn: function names must be symbols")))
    (list form name (list (sym :fn) args ...)))
```

technomancy:
@bb010g yeah, that's more or less what I had in mind.

thatismatt:
@bb010g thanks. That's really helpful. I've updated `defn` to now cater for multi-syms too.

PR #59, opened 2018-04-06
When macro

technomancy:
I started converting `when` over to a macro, and it caused some test failures because the error message didn't have proper line numbering. I think it's necessary to expose `assertCompile` to Fennel code in compiler scope. However, just exposing the function itself isn't enough; the macros also need the ast node, because that's where the actual macro name and line number info is stored.

I started trying to find a way to expose this, but it was getting kind of out of control. The problem is that the ast node can't be exposed in `makeCompilerEnv` because it's not known yet, so you have to find another way to expose it to the macro. My first attempt was to pass it as the first argument to the macro, but that's counter-intuitive; no one expects the macro to get more arguments than it's called with.

What I ended up doing was exposing an `assert-compile` function into the macro's environment table using `getfenv`. But the `assert-compile` I expose is a wrapper around the regular `assertCompile` which automatically passes in the `ast` argument in the right place, so the macro author doesn't have to think about it; `assert-compile` just takes the same args as `assert`. (Maybe it should be renamed to just `assert`?) 

Unfortunately to do this I had to add a port of `getfenv` which works on 5.2+, so I'm not thrilled with this approach. Definitely interested in hearing alternative ideas if you've got any.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -1153,18 +1153,6 @@ SPECIALS['if'] = function(ast, scope, parent, opts)
     end
 end
 
--- (when condition body...) => []
-SPECIALS['when'] = function(ast, scope, parent, opts)
-    assertCompile(#ast > 2, 'expected body', ast)
-    table.remove(ast, 1)
-    local condition = table.remove(ast, 1)
-    local body = list(sym("do", ast[1].line, ast[1].filename), unpack(ast))
-    local new_ast = list(sym("if"), condition, body)
-    new_ast.line, body.line = ast.line, ast.line
-    new_ast.filename, body.filename = ast.filename, ast.filename
-    return SPECIALS["if"](new_ast, scope, parent, opts)
-end
-
 -- (each [k v (pairs t)] body...) => []
 SPECIALS['each'] = function(ast, scope, parent)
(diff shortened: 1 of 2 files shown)


bakpakin:
Interesting thoughts, but would you want assert-compile to use the ast that was passed as an argument to the macro invocation, not the macro that was returned? In the case of when, the `(when ...)` form that was parsed exists even if the when macros itself fails.

I think we can simply pre-bind assert-compile to use this as the default option for ast. Am I missing something? I don't see why we need getfenv.

bakpakin:
Perhaps we could modify macroToSpecial by wrapping it with a pcall to catch any errors. If we catch errors, we could prepend the ast line information, and the rethrow. This way, the macro author can just use assert. We would also catch generic errors in the macro and report them correctly.

technomancy:
Ooooooh yeah `pcall` is a much better solution here. Great idea; I'll switch it up to do that.

technomancy:
I got rid of the bit where we tweak the error messages and changed `assertCompile` to leave out the line number of the `assertCompile` call itself since that was just useless noise at the beginning of the line:

```
Compile error in `defn1' in.fnl:3: [string "./test-macros.fnl"]:11: defn: function names must be symbols
```

Issue #53, opened 2018-04-04
Improve cosmetics of compilation output


technomancy:
A few things that would be nicer:

* [x] Fixing a bug with side-effects in 9acf9cda88be5e1071b2d1639d016a7651fbbf8b caused spurious `do _ = nil end` lines to be emitted in compilation output.
* [ ] Functions are always compiled and set to gensym'd locals; this is a bit untidy when they could be emitted directly as expressions.
* [x] Can we detect cases where mangling dashes to `_2d` is unnecessary? If dashes could mangle to underscores when there's no conflict then the code emitted would be a fair bit more readable.

bb010g:
Regarding mangling dashes, I'd actually be up for doing a Nim-style mangling where dashes are always converted into underscores. We can't tell if an arbitrary table `t` has a key `a_b` without testing every time anyways, and would make dealing with Lua libraries that use snake case consistently much nicer.

Benaiah:
@bb010g that seems really confusing to me, considering that you can use underscores literally in Fennel already. `(let [a-b 1 a_b 2] a-b)` should _not_ return 2 IMO. Additionally, that approach doesn't do anything for CamelCase, which in my (limited) experience is more common than snake_case.

The biggest issue with dash-mangling that I see is that we have no way of knowing what's in the global table at runtime, so there's no way to detect conflicts with those names.

bakpakin:
Yes, there are some issues. Namely, if you create a global `my-global`, it can be accessed with `my_global`, which is a bit strange.

@Benaiah your example will work fine regardless. The current name mangler treats local variables a bit differently than globals, and will automatically change the mangling to be unique for every local variable.

bb010g:
@Benaiah Yeah. The biggest issue is that you'd break existing code that used both dashes and underscores like that right now, but pre-1.0 Fennel is the best time to do stuff like that. And you're right about it not doing anything for CamelCase, but that's because we can't control Lua's semantics, unlike Nim. I just can't think of many cases where you'd actually want to have both `a-b` and `a_b` as variables, and this is a case where we can improve readability without running into Lua constraints. If we make it clear that dashes in symbols are really underscores, I don't think there would be any problems with it after the first time someone runs into it.

It would be interesting though to see if you could use a macro with `eval-compiler` to change the mangler and swap a portion of code over to mangling snake-case as camelCase, which you could then disable when done.

Benaiah:
[...]

PR #60, opened 2018-04-06
Clean output

bakpakin:
This is an attempt to begin addressing a number of issues, including #53 and source mapping issues.
It currently fixes the `do local _ = ... end` bug from #53, adds more granularity to source mapping (adds byte offsets as well as line information) by embedding mapping information in comments at the ends of lines, and removes empty lines from compilation output. I would like to also like to address mangling here,
removing excessive parens in the output, and restructuring the if form to not generate such deeply nested if elses, and uses the native lua `elseif`.


bakpakin:
The current output is much nicer now. To see the new output without sourcemaps, try `./fennel --compile --sourcemap false generate.fnl`. If you remove the --sourcemap option, you can see sourcemap info is appended to each, line, with source line, source byte offset start, and source byte offset end. I have also added the dash to underscore conversion for nicer output.

fennel.lua
```diff
@@ -524,16 +545,15 @@ local function exprs1(exprs)
...
-    table.insert(chunk, {out, line = ast and ast.line})
-end
-
 -- Compile sideffects for a chunk
 local function keepSideEffects(exprs, chunk, start, ast)
     start = start or 1
     for j = start, #exprs do
         local se = exprs[j]
-        if se.type == 'expression' then
+        -- Avoid the rouge 'nil' expression (nil is usually a literal,
```
Benaiah: ITYM `rogue 'nil' expression`.

fennel.lua
```diff
@@ -401,12 +424,14 @@ end
...
-    if luaKeywords[symb] then
+    if luaKeywords[symb] or symb:match('^%d') then
         symb = '_' .. symb
     end
-    return symb:gsub('[^%w_]', function(c)
+    local function escaper(c)
         return ('_%02x'):format(c:byte())
-    end)
+    end
+    symb = symb:gsub('-', '_')
```
Benaiah: Does this cause dashes to always be converted to underscores? Copying my concerns from https://github.com/bakpakin/Fennel/issues/53#issuecomment-379371973:

> ...that seems really confusing to me, considering that you can use underscores literally in Fennel already. `(let [a-b 1 a_b 2] a-b)` should return `1` IMO. Additionally, that approach doesn't do anything for CamelCase, which in my (limited) experience is more common than snake_case.

fennel.lua
```diff
@@ -401,12 +424,14 @@ end
 
 -- Creates Lua-friendly symbol from a single symbol by mangling it.
 local function symMangle(symb)
-    if luaKeywords[symb] then
+    if luaKeywords[symb] or symb:match('^%d') then
         symb = '_' .. symb
     end
-    return symb:gsub('[^%w_]', function(c)
+    local function escaper(c)
```
Benaiah: Is there a particular reason to create this function within `symMangle` rather than in the scope above it?

bakpakin: I have moved it back to how it was before as it was clearer.

technomancy:
Great to see this coming together. I like the `elseif` changes; that
makes a big difference in readability. In fact, at this point really
my only remaining readability concern is the `_1_`, etc locals created
for every function.

Out of curiosity, what do you imagine the source maps being used for?
Is there a way you can feed them to a Lua implementation so that it
will use that data for stack traces? It seems like this removes the
ability for the user to get meaningful correspondence in their stack
traces, so the idea of removing that before an alternative method
is ready feels a bit concerning.

I saw that you got rid of `macro` but kept `special`. Is `special`
redundant if folks can do stuff like `(eval-compiler (tset _SPECIALS
:myspecial (fn [...] ...)))`? I could see an argument being made for not wanting to commit to exposing `_SPECIALS`.

bakpakin:
We actually can feed sourcemaps to Lua to in a custom error handler via the xpcall. We could then use the debug library (debug.traceback) to get the Lua stack such and remap each line in the stack trace to the fennel source.

bakpakin:
And no, special is not special :). I think we can also get rid of luaexpr and luastatment as well. The compiler environment is sufficiently powerful already it seems.

bb010g:
I currently require `luastatement` to get out of an IIFE in a hot loop. Providing these escape hatches is probably a good thing long-term, even if they're almost always not necessary.

As for getting rid of the `_1_` locals, I want to try my hand at getting rid of those in a PR after this is merged.

bakpakin:
I've removed the `special` special form, but left in the luaexpr and luastatement specials for now. As @technomancy pointed out, eval-compiler combined with tset _SPECIALS does much the same thing.

technomancy:
Well, I don't love the idea of requiring a special error handler before line numbers make sense, but I'll hold my judgement till I see how it works in practice.

bakpakin:
These are some rather large commits as they change the mangling a bit, and add more meta data on bindings (basically, extending the current var tracking). Also a lot of work on updating destructure to be more strict about what it accepts for bindings.

For example, global, local, var, and let will now not accept multisyms. You must use set if you want to update a multisym.

There is now a difference between the local name mangling, which is internal to the compiler, and the global name mangling, which is intended to be used as an interface to Fennel if needed. A `fennel.mangle` and `fennel.unmangle` function are made available in the API.

This perhaps should have gone in a separate PR, but it depended on the changes from this branch.

bakpakin:
This latest commit adds the `fennel.traceback` function that allows proper error handling. I ended up doing something similar to [what leafo did with moonscript](https://moonscript.org/reference/command_line.html#command-line-tools/moon/error-rewriting), and just optionally store the source mapping in memory. This removes the ugly comments add the end of output, but means there is no source mapping for fennel that has been compiled to Lua ahead of time (I suppose maps could be stored in .map file like JS and loaded into memory manually). It works by default in the repl and the fennel script, and usage when using fennel from Lua is something like

```lua
local ok, res = xpcall(function ()
  return fennel.dofile('mycode.fnl')
end, fennel.traceback)
if not ok then
  -- handle the stacktrace in res
  print(res)
else
  -- handle the result in res
  print('OK! ' .. tostring(res))
end
```

The resulting stacktraces look much like Lua stacktraces, but the file and line information for fennel code should be correct, including code run via `fennel.eval`.

Example error producing fennel:
```lisp
;; this will error because x will be :woof
(local bork (fn [x] (+ x 1)))

(local bark (fn [x]
             (bork x)
             nil))

(local dog (fn [x]
             (bark x)
             :cat))

(dog :woof)

{}
```

Then run with fennel command, fennel.dofile, in the repl, or fennel.eval.
```
./fennel error.fnl
```

Should print

```
fennel error: attempt to perform arithmetic on local 'x' (a string value)
stack traceback:
  error.fnl:2: in function 'bork'
  error.fnl:5: in function 'bark'
  error.fnl:9: in function 'dog'
  error.fnl:12: in main chunk
  (tail call)
  (tail call)
  (tail call)
  [C]: in function 'xpcall'
  ./fennel:21: in function 'dosafe'
  ./fennel:75: in main chunk
  [C]: in ?
```

technomancy:
Interesting approach with the traceback; seems to work pretty well!

It looks like there's currently a limitation that the traceback must be generated by the same process as the one that initially compiled the code, but this doesn't necessarily need to be the case moving forward as source maps could be saved and propagated to runtime?

Also, doesn't this mean that including the source map comments in the compiler output isn't actually helpful for AOT compilation as they will never actually get used?

bakpakin:
> Also, doesn't this mean that including the source map comments in the compiler output isn't actually helpful for AOT compilation as they will never actually get used?

Yeah, I have removed the end-of-line comments for now. In the future, it could be possible to save source maps from compiled code inline in Lua comments, or in a separate file. I'm liking the latter right now. Some function like `fennel.loadmap(luafile, mapfile)` could then load the sourcemap into memory.

PR #65, opened 2018-04-13
[fennelview] Be more permissive in the map keys that are printed as keywords.

thatismatt:
I've only added `- _ > <` which is less permissive than then fennel parser. If people think we should allow more characters here then I'm happy to extend this PR to include more characters.

My reasoning for not extending to the full range of fennel keywords was that I wondered if there was an implicit heuristic going on here, e.g. if a keyword/string includes certain characters then it could've been written as a keyword but it is more idiomatic to write it as a string.

### DIFF
--- a/fennelview.fnl
+++ b/fennelview.fnl
@@ -102,7 +102,8 @@
                               (puts self " ]")))
 
 (local put-key (fn [self k]
-                 (if (and (= (type k) "string") (not (: k :find "%W")))
+                 (if (and (= (type k) "string")
+                          (: k :find "^[%w?\\^_`!#$%&*+-./@~:|<=>]+$"))
                      (puts self ":" k)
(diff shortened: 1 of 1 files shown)


bakpakin:
> If people think we should allow more characters here then I'm happy to extend this PR to include more characters.

Yeah, we might as well. Anything that can be written as a keyword should be printed such.  The `issymbolchar` function in the parser determines which characters are acceptable in both a symbol and a keyword.

thatismatt:
Are you thinking we should add a dependency from fennelview on fennel?

bakpakin:
No, its a very simple function and can be made into a Lua pattern to match strings instead of characters.

thatismatt:
Re-pushed, this now matches issymbolchar from fennel.lua.

PR #66, opened 2018-04-14
Fix if soundness

bb010g:
While updating [otouto](https://github.com/topkecleon/otouto)'s vendored Fennel, I found that this Fennel code
```
(if (: (a) :a)
    1
    (: (b) :b)
    2
    ; else
    3)
```
compiles to this Lua
```
  local _0_ = a()
if _0_[("a")](_0_) then
  return 1
  local _1_ = b()
elseif _1_[("b")](_1_) then
  return 2
else
  return 3
end
```
which is unsound, as the temporary variable `_1_` that holds the result of `(b)` is created under the previous branch as dead code.

This is fixed by keeping track of whether there's a condchunk for a conditional branch. There aren't any tests for this yet, but I wanted to get the PR up quickly.

---
This change is [Reviewable](https://reviewable.io/reviews/bakpakin/fennel/66)

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -1147,19 +1147,25 @@ SPECIALS['if'] = function(ast, scope, parent, opts)
     -- Emit code
     local s = gensym(scope)
     local buffer = {}
+    local lastBuffer = buffer
     for i = 1, #branches do
         local branch = branches[i]
-        local fstr = i == 1 and 'if %s then' or 'elseif %s then'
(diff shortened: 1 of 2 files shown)


bb010g:
Ok, a test has been rebased in that passes on both commits.

bakpakin:
Ah thanks. I remember now why I added the nested if else.

PR #67, opened 2018-04-14
Beautify output

bb010g:
Removes extra parens around strings, makes the `.` form use `.ident` when possible, makes the `:` form use `:method` when possible, and makes table syntax use `{ a = 1 }` instead of `{ ["a"] = 1 }` when possible.

All the tests I've ran seem to show that the string parens were unnecessary, and I couldn't find anything in the commit log or issues talking about them. The only case where literals have to be in parens is for method calls, and we're already handling those.

---
This change is [Reviewable](https://reviewable.io/reviews/bakpakin/fennel/67)

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -387,7 +387,7 @@ end
 
 -- Allow printing a string to Lua
 local function serializeString(str)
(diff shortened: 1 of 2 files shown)


bakpakin:
Awesome, thanks!

PR #73, opened 2018-05-15
Macro-oriented changes

bb010g:
I ran into the need for `set-forcably!` when implementing a macro that did default arguments. Argument bindings aren't mutable by default, and they shouldn't be if you're just doing a default argument to maintain the same feel, so you need a way to bypass the normal `var` check.

The other change is just nice for compile times in projects that require complex macro modules a lot.

---
This change is [Reviewable](https://reviewable.io/reviews/bakpakin/fennel/73)

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -827,6 +827,7 @@ local function destructure(to, from, ast, scope, parent, opts)
     local nomulti = opts.nomulti
     local noundef = opts.noundef
     local forceglobal = opts.forceglobal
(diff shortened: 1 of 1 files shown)


bakpakin:
Looks good, also caching the required macro modules.

laheadle:
You misspelled 'forcibly'.

bb010g:
Thanks for catching that. Guess I'll drop a PR and add some tests while I'm at it. :D

PR #74, opened 2018-05-21
Reintroduce line number correlation feature.

technomancy:
I need this to integrate Fennel into TIC-80 because it calls the C
implementation of debug.traceback directly rather than allowing it to
be overridden from within Lua code.

I've also introduced make_searcher so you can override the options
passed to the compiler when it's invoked by require.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -546,6 +546,33 @@ local function peephole(chunk)
     return chunk
 end
 
+-- correlate line numbers in input with line numbers in output
+local function flattenChunkCorrelated(mainChunk)
+    local function flatten(chunk, out, lastLine, file)
+        if chunk.leaf then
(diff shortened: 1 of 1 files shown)


bakpakin:
Given the desire for line numbers without a custom debug and no reason not to, I will merge this back in. I am still skeptical of how well this works, however, especially if the programmer puts more than one expression per line, and the source code "lags" behind the emitted code. Also the problem you have mentioned before about macros and reordering, but I think that seems uncommon with most macros.

Also, is there any reason to not to emit well indented Lua? Sure, it's meant to be run by the interpreter and not written by the programmer, but I don't see why we shouldn't indent the code well either. When things go wrong, the generated Lua itself can be a valuable debugging tool.

bakpakin:
As has been pointed out on IRC, the line correlation feature would be degraded if proper indentation was forced (either that or the indentation would be strange), so nixing that.

PR #79, opened 2018-06-13
Catch typos that aren't in allowed globals list.

### LINKED ISSUE #77: Prevent typo'd global references
We already prevent accidental global setting with `set`, but there's currently no way to prevent typos from accidentally referring to a global. I would like to add a strict mode which would prevent this from happening unless  the it's in a list of allowed globals.

This is pretty simple to do with one catch: passing the list of globals thru from the compiler entry point reliably to `compile1` which needs to check it. Originally I thought the `opts` map would be the right place for this, but this table seems to be used exclusively for compiler internals, and storing a user-provided value in there not only feels wrong but doesn't really work anyway since `opts` tables are frequently replaced and not propagated downstream.

As far as I can tell the only way to support this would be to introduce a new top-level mutable local for user-provided compiler settings or to start storing those settings in the existing `opts` table which is already an argument to many functions, and modify all those calls to propagate the user-provided settings thru. Since there are over 30 places this happens that would need to be modified, I'm inclined towards the top-level mutable local, but maybe there's another way I haven't thought of.

technomancy:
Note that this uses a mutable top-level local variable, because the
alternative discussed in #77 is very invasive.

This is off by default unless you pass allowedGlobals in the options
table when invoking the compiler. In the future it may be possible to
seed it with the default set of keys in _G, but that will require more
discussion.

Fixes #77.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -653,6 +653,18 @@ local function exprs1(exprs)
     return table.concat(t, ', ')
 end
 
+-- If there's a whitelist of allowed globals, don't let references
+-- thru that aren't on the list. This list is set at the compiler
+-- entry points of compile and compileStream.
+local allowedGlobals
+
+local function globalAllowed(name)
+    if not allowedGlobals then return true end
+    for _, g in ipairs(allowedGlobals) do
+        if g == name then return true end
+    end
+end
+
 -- Compile side effects for a chunk
 local function keepSideEffects(exprs, chunk, start, ast)
     start = start or 1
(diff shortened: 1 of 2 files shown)


technomancy:
Thanks. The main potential problems I see with the top-level variable changing is if we add more entry points into the compiler, (this could be helped if we made it so `compile` was the sole entry point for the compiler and implemented `compileStream` in terms of it) or if the compiler could be somehow in the middle of compiling two codebases at the same time (using coroutines or something). But I think this is worth it.

technomancy:
Talking in chat, it feels like there are two distinct cases where this could come into play. For ahead-of-time compilation you can't assume it's safe to turn this on; you're likely to be compiling in an instance of Lua that has a completely different set of globals than the one you're targeting. But in cases of "live" compilation (`eval`, `dofile`, and `require`) you're almost always safe assuming that the set of globals available at compile time is the same as the set you'd use at runtime, right?

So do you think it would make sense to seed `allowedGlobals` inside those calls with the keys present in `_G`? Of course we'd allow it to be turned off if `allowedGlobals` were set to false.

PR #84, opened 2018-09-12
Accept --searcher flag to install package.searchers entry.

technomancy:
This lets you use Fennel files that `require` other Fennel modules without cluttering your own code with the `package.searchers` insertion; nice for scripts and stuff.

I'm not thrilled with this ad-hoc method of checking args. It works for now but should probably be overhauled when we add another arg.

Also, maybe during repl invocations and "run this file" invocations the searcher should default to being included, and we could have a --no-searcher argument instead. Thoughts?

### DIFF
--- a/fennel
+++ b/fennel
@@ -10,6 +10,7 @@ Usage: fennel [FLAG] [FILE]
   --repl          :  Launch an interactive repl session
   --compile FILES :  Compile files and write their Lua to stdout
   --help          :  Display this text
+  --no-searcher   :  Skip installing package.searchers entry
 
   When not given a flag, runs the file given as the first argument.]]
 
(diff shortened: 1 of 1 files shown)


bakpakin:
Looks good to me, and I think it should definitely be  the default.

> I'm not thrilled with this ad-hoc method of checking args. It works for now but should probably be overhauled when we add another arg.

That's a good point. It shouldn't be too hard to roll up the various flags into a more uniform mechanism, but if it works for now definitely wait.

technomancy:
Cool; thanks. Pushed a change to turn it on by default. Wasn't too much work to unify the option flags so I switched to handling them uniformly.

PR #86, opened 2018-09-14
Add destructuring for each and fn bindings.

### LINKED ISSUE #85: Destructuring fn, each locals?
Right now destructuring only applies to `local`, `let`, `var`, `global`, and `set`. It would be convenient if locals introduced by `fn` could also have destructuring applied. @bakpakin is this something you were planning on implementing, or was there a reason to leave it out?

technomancy:
It works fine, but the output is somewhat less than ideal as it
introduces one level of unnecessary binding:

```clj
(each [_ [a b] (ipairs [[1 2] [3 4]])]
  (print (+ a b)))
```

```lua
for _, _0_0 in ipairs({{1, 2}, {3, 4}}) do
  local _1_ = _0_0 -- <- this bit here is redundant
  local a = _1_[1]
  local b = _1_[2]
  print((a + b))
end
```

I'm probably abusing `declareLocal` here; any advice on how I could
clean that up?

Fixes #85

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -1083,10 +1083,14 @@ SPECIALS['fn'] = function(ast, scope, parent)
             assertCompile(i == #argList, "expected vararg in last parameter position", ast)
             argNameList[i] = '...'
             fScope.vararg = true
-        elseif isSym(argList[i])
-            and argList[i][1] ~= "nil"
-            and not isMultiSym(argList[i][1]) then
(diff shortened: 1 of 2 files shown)


bakpakin:
Yeah, looking at the diff it seems this could be simpler. I'll take a look at this in a bit.

technomancy:
Maybe we can get this into 0.1.0 as-is and come back to clean up the output later.

PR #92, opened 2018-09-26
Allow globals from an environment when one is provided.

m-r-hunt:
When no `allowedGlobals` list is provided, the current behaviour is to use a list based on the globals in the current environment. This is confusing when the user has supplied an environment for the code to run in, as the globals from that environment won't be allowed by the compiler. This is compounded by the fact the `allowedGlobals` mechanism isn't documented.

I've made a change so that the `allowedGlobals` list is instead generated based on the environment the code is running in. If none is supplied, it works as it currently does. 

I ran `make ci` without any problems, and these changes are quite simple. Let me know if you have any feedback.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -1577,9 +1577,10 @@ local function traceback(msg, start)
     return table.concat(lines, '\n')
 end
 
-local function currentGlobalNames()
+local function currentGlobalNames(env)
+    env = env or _G
     local names = {}
-    for k in pairs(_G) do table.insert(names, k) end
+    for k in pairs(env) do table.insert(names, k) end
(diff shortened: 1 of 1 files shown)


Benaiah:
I agree that the behavior you describe is a much nicer default. I ran into this issue when introducing a repl in my project.

Haven’t tested it yet, but it LGTM at a first glance though some comments need to be fixed since the behavior they describe has changed.

technomancy:
Thanks for catching this oversight. Your fix makes a lot of sense to me. I dropped the ball on getting this documented, so I apologize for the confusion there.

Issue #100, opened 2018-10-22
Macros in the repl


cosarara:
What I tried:
```
$ fennel --repl
Welcome to fennel!
>> ((lambda [] (require-macros "test-macros") (->1 9 (+ 2) (* 11))))
Compile error: Compile error in 'ipairs' ./test-macros.fnl:4: unknown global in strict mode: ipairs
```

What should happen:
```
$ fennel --repl
Welcome to fennel!
>> ((lambda [] (require-macros "test-macros") (->1 9 (+ 2) (* 11))))
121
```

technomancy:
Fixed by #103

PR #103, opened 2018-10-29
Fix globals for Fennel macros

bb010g:
Ran into this while trying to compile macro-using files from the command line.

---
This change is [Reviewable](https://reviewable.io/reviews/bakpakin/fennel/103)


FongHou:
Is the a proper fix for #100? I tried the patch, it seems work as expected.

>> (do (require-macros :test-macros) (->1 9 (+ 2) (* 11)))
121

technomancy:
This looks great; thanks! I think it does take care of #100.

PR #109, opened 2018-11-08
Add changelog up to version 0.1.0.

technomancy:
Anything else that should go into 0.1.0 before we call it good?


bakpakin:
Looks good for now, that’s quite a lot already.

PR #110, opened 2018-11-15
Allow stdin with --compile

thatismatt:
A tiny change to add the ability to compile Fennel code supplied on stdin to the `fennel` command.

Usage would be:
```
$ echo "(print (+ 1 2))" | fennel --compile -
return print((1 + 2))
```
or:
```
$ echo "(print :file-a)" > a.fnl
$ echo "(print :file-b)" > b.fnl
$ echo "(print :stdin)" | fennel --compile a.fnl - b.fnl 
return print("file-a")
return print("stdin")
return print("file-b")
```


bakpakin:
Looks good!

PR #111, opened 2018-11-17
Remove defn.

technomancy:
I think we can get rid of this; if I had known that `fn` could be used the same way as `defn` then I would not have added it in the first place. Figure we should get all our breaking changes in before 0.1.0.

Any thoughts?

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,7 @@
 
 ## 0.1.0 / ??
 
+* Remove `defn` macro as it is redundant.
 * Add `doto` macro
 * Support newlines in strings
 * Prevent typos from accidentally referring to unknown globals
--- a/fennel.lua
+++ b/fennel.lua
@@ -1877,9 +1877,6 @@ local stdmacros = [===[
              (table.insert form elt))
            (table.insert form name)
            form))
- :defn (fn [name args ...]
-         (assert (sym? name) "defn: function names must be symbols")
-         (list (sym :fn) name args ...))
  :when (fn [condition body1 ...]
          (assert body1 "expected body")
          (list (sym 'if') condition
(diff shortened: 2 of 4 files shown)


bakpakin:
It is redundant. I suppose it could have had extra functionality like the lambda macro, but I’m all for slimming things down.

kriyative:
@technomancy @bakpakin - actually, I would humbly suggest leaving `defn` in, for at least a couple of reasons:
- it's already useful as a way to apply different font-lock rules in Emacs
- I'm experimenting with alternative module/namespace strategies and it could be useful for that

Thanks.

technomancy:
> * I'm experimenting with alternative module/namespace strategies and it could be useful for that 

If it becomes part of an alternative module system, shouldn't it ship as part of that module system instead of being included in its current state and having it mean something new later?

kriyative:
@technomancy - I don't think I need (or want) to change the semantics of `defn` in any way. Having it would be cheap way to handle top level function definitions differently from inner ones. While I understand that current implementation makes no real distinction between top-level and inner fns in the generated Lua, seems like an inexpensive thing to keep - especially since it's already there!

Benaiah:
The issue I have with the current `defn` is that `defn` is a very useful term, and making it an alias for something that `fn` already handles seems like a bit of a waste. Therefore, I'm in favor of removing it for now - though perhaps we should consider reserving the name (at least by default) in the compiler so it can be used without breaking backwards compatibility in the future.

technomancy:
I wonder if the changes in #114 would make the current behavior of `fn` more repl-friendly and reload-friendly to a degree that augmenting it with `defn` is no longer needed? Please take a look.

technomancy:
I'm going to go ahead and get rid of this. I'm not averse to reintroducing it in the future if there's more going on than just an alias, but in the mean time a first-class module system could probably be evolved as a macro first rather than being part of the core language.

PR #113, opened 2018-11-21
Clarify optional arguments usage in lambda

### LINKED ISSUE #112: Unexpected behaviour for `lambda`
The tutorial mentions this code:
```
(global print-calculation
        (lambda [x ?y z] (print (- x (* (or ?y 1) z)))))
(print-calculation 5) ; -> error: Missing argument z
```
I had expected that this would mean that you could invoke this function with only two arguments, but that isn't the case:
```
(print-calculation 5 6) ; -> error: missing argument z
```

I think it was this sentence from the tutorial that confused me:
> Note that the second argument `?y` is allowed to be omitted, but `z` is not.

Is the expected usage more like this:
```
(print-calculation 5 nil 6)
```

Is this clarification that is required in the documentation or a bug in `lambda`?

thatismatt:
Fixes #112

### DIFF
--- a/tutorial.md
+++ b/tutorial.md
@@ -46,7 +46,11 @@ an error.) For safer code you can use `lambda`:
 (print-calculation 5) ; -> error: Missing argument z
 ```
 
-Note that the second argument `?y` is allowed to be omitted, but `z` is not.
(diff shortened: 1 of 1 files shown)


technomancy:
Thanks!

PR #114, opened 2018-11-22
Implement saveLocals feature of the repl.

technomancy:
This implements a feature whereby locals in the repl are preserved
from one chunk to the next, which is currently one of the biggest
annoyances in the Fennel repl.

For every compilation, when the saveLocals isn't false, the repl will
compile the input, then in the penultimate line, it will inject a
snippet of code which loops over all the top-level locals in the given
chunk using the debug.getlocal function, and saves them off to a
table. After the code has been evaluated, all the values which have
been stored in this table get spliced into the current environment.

This more or less gives you a repl situation that acts "the way you'd
expect", allowing fn declarations from one input to remain around for
subsequent inputs.

The drawbacks are such:

* It introduces a ___replLocals___ global into the environment in
  order to have a place to store the values.

* Any local set in a chunk essentially becomes a global for the
  purposes of subsequent calls. This could have unintended consequences.

* It relies on assumptions about the string output of the compiler,
  namely that the last line is always a return, and that it's safe to
  inject code into the line right before the last line.

I think the first and last issues are relatively minor and worth the trade-off. The second concern I'm less confident about--I feel like this could result in some sticky situations. Maybe it would be better to just splice the the locals by prepending a `local name = ___replLocals___["name"]` line to each evaluation instead?


bakpakin:
Looks good to me. This will definitely make the repl more useful and less surprising. One thing that maybe could be better is the binding to __i__  could replaced by a gensym.

As for the assumptions about the compiler, I think this a good, easy way to get locals that should work in most if not all real cases. Another possibility is to just make top level locals into globals.

technomancy:
Oh crap; just realized this doesn't work with mangled names. `=\`

Issue #96, opened 2018-10-06
Confusing interaction between parse errors and multiline input at the REPL


thatismatt:
I was confused earlier by the fact that a parse error could put the REPL in a state where it appears that no output is being printed. Here is an example REPL session, I have marked the moment of my confusion!

```
Welcome to fennel!
>> (1x 1)
Parse error: could not read token "1x" in unknown:1
>> 1  ;; <--- confusion here
>> )
Compile error: Compile error in '1' unknown:1: cannot call literal value
>> 1
1
```

I'm not sure what the fix is as I think there are actually two issues:
1. Parse errors should probably "reset" the parser state, in particular I think the issue in this case is that the stack isn't reset.
2. Multiline input "mode" at the REPL is indistinguishable from the first line of input, for reference Lua doesn't appear to allow multiline input, Python changes from `>>>` to `...`, Clojure changes from `namespace=>` to `  #=>`. (I had a quick look as to whether this could easily be added to the Fennel REPL, but the point that the prompt is printed doesn't currently have access to the parser state.)

bakpakin:
Yes, this is a good point and should be relatively straight forward. I have implemented exactly this functionality before with such a parser and getting python or clojure functionality is a matter of exposing the parser state as you said. The parser should also definitely reset the stack on an error.

technomancy:
Nice work getting that PR in. What's remaining before we can consider this closed?

thatismatt:
I think the parser stack being reset on errors makes sense, and I think that is an easy fix. I can take a look over the weekend.

For reference, below is the new behaviour when a parse error occurs:

```
>> (1x)
Parse error: could not read token "1x" in unknown:1
Compile error: Compile error in '()' unknown:1: expected a function to call
>> (1x
Parse error: could not read token "1x" in unknown:1
.. )
Compile error: Compile error in '()' unknown:1: expected a function to call
>> (1x )
Parse error: could not read token "1x" in unknown:1
.. )
Compile error: Compile error in '()' unknown:1: expected a function to call
```

Each of these cases seem odd to me. I think it boils down to what the expected behaviour is after the parse error. i.e. should it carry on and get to the compilation stage.

bakpakin:
In general, recovering from a parse error is probably quite difficult and there probably isn’t a good way to get “close” to what the user wanted. The way that makes sense to me is to totally clear the parse stack (all open parens) and reset the parser to the root state.

The problem with this is that there will be a parse error for every following closing delimiter.

[...]

PR #122, opened 2018-12-07
Display .. instead of >> at the repl, when more input is expected.

thatismatt:
This PR threads the parser state through from the parser to the read chunk function, where if the parser stack is empty it will display `>>` otherwise it displays `..`. The parser state is a table, but it only contains the parser's stack size at the moment, but it could be extended in the future.

This addresses some of the issues in #96 .

Here is some example output to show what this change achieves:

```
$ fennel
Welcome to fennel!
>> (+
.. (+
.. 1
.. 2)
.. 3)
6
>> 
```

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -119,14 +119,14 @@ local function granulate(getchunk)
     local c = ''
     local index = 1
     local done = false
-    return function ()
+    return function (parserState)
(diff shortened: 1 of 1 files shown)


bakpakin:
Nice, looks great!

PR #125, opened 2018-12-21
quote

noisesmith:
This is a first pass at implementing quote as a special form for `

unquote is not implemented but there's a clearly marked TODO for where it would go

I expect there are multiple things here that could be done more elegantly - I'll be focusing on another project for a little while, others are welcome to improve or finish up this feature for merge.

I am of course willing to answer questions about what I'm trying to do here.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -344,7 +344,7 @@ local function parser(getbyte, filename)
                 end
             end
         until done
-        return true, retval, lastb, line, byteindex
+        return true, retval
     end, function ()
         stack = {}
     end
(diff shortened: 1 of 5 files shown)


noisesmith:
a loose end on this PR: it emits calls to `sym` and `list` for obvious reasons, but doesn't yet add sym and list to the scope of normal fennel code - perhaps given that ` is mostly useful in macros this is a non issue though(?)

noisesmith:
The unquote feature is implemented as well, using `~` but special casing `~=` since that function is ubiquitous including in the compiler internals

bakpakin:
Also, perhaps we could use the @ character for unquote to avoid the collision with ~=? I don't think @ is currently used for anything so it would most likely be backwards compatible.

fennel.lua
```diff
@@ -1529,6 +1543,86 @@ local function compile(ast, options)
...
+    end
+    return ret
+end
+
+-- expand a quoted form into a data literal, evaluating unquote
+local function do_quote (form, scope, parent)
+    local q = function (x) return do_quote(x, scope, parent) end
+    -- symbol
+    if isSym(form) then
+        return ("sym('%s')"):format(deref(form))
```
technomancy: The `tostring` behavior of symbols should make `deref` redundant here.

fennel.lua
```diff
@@ -1529,6 +1543,86 @@ local function compile(ast, options)
...
+        if not(seen[k]) then
+            ret = ret .. s .. '[' .. k .. ']' .. '=' .. v
+            s = joiner
+        end
+    end
+    return ret
+end
+
+-- expand a quoted form into a data literal, evaluating unquote
+local function do_quote (form, scope, parent)
```
technomancy: I haven't been terribly consistent about this, but the compiler ought to use camelCase for its functions here.

noisesmith:
I'll be able to address most of this today, thanks for the feedback.

technomancy:
I played with this and was able to get it to track when quoting lists or symbols is allowed: https://github.com/bakpakin/Fennel/commit/396b1eeebf78e1a609fed782097e0a4707ce95e1

@bakpakin interested in your thoughts on this; should we explicitly block runtime quoting of lists and symbols?

bakpakin:
I think it's probably good to disable quoting at runtime. Fennel cannot rely on any runtime machinery besides what is provided by Lua, and the errors generated were confusing.

PR #126, opened 2018-12-21
Implement pattern matching macro.

technomancy:
Tried to give an overview of how pattern matching works in `reference.md` but I'm not sure how well I explained it; suggestions for improving the clarity are welcome. The test suite has a more thorough list of supported features with examples.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,7 +1,8 @@
 # Summary of user-visible changes
 
-## 0.1.2 / ???
+## 0.2.0 / ???
 
+* Add `match` macro for pattern matching
 * Fix a bug where runtime errors were not reported by launcher correctly
 * Allow repl to recover gracefully from parse errors
 
--- a/fennel.lua
+++ b/fennel.lua
@@ -1815,7 +1815,7 @@ local function searchModule(modulename)
     end
 end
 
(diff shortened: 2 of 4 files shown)


technomancy:
This still needs work; it gives the wrong result with this:

```lisp
(match [1 2 3] [x y x] :no [x y z] :yes)
```

Needs to do intra-pattern unification, not just unification with existing locals.

bakpakin:
I have had a look at the code, and I think supporting intra-pattern unification can be done by keeping a table of symbols seen already. If a symbol has been seen already while traversing the pattern, we instead check if the value is equal to the previous binding instead of generating a new binding.

This would require that _all_ previous bindings are available to subsequent bindings. Perhaps declaring all bindings up front with local and using set-forcibly! to add bindings would do the trick.

technomancy:
This can be done without set-forcibly!; it's just a new table we can put things in. I've pushed an implementation; you can see the changes to the tests to see how it works.

I also added a special case to allow `_` to be a wildcard which always matches.

I think this is probably ready to merge now; what do you think?

bakpakin:
Perfect, I like the wildcard! I guess set-forcibly wasn't needed at all.

PR #127, opened 2018-12-21
Add -?> & -?>> AKA the thread maybe macros.

thatismatt:
This transforms the threaded forms in to nested ifs, which has the advantage of early exit. But the disadvantage of potentially creating a deeply nested Fennel AST & Lua code.

We could (and I notice that this is what Clojure's `some->` does) transform this in to a "flat" let instead.

If you like this approach I'll tweak it for the "thread last maybe" `-?>>` macro. If not, I'm happy to rewrite it.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,7 @@
 
 ## 0.2.0 / ???
 
+* Add `-?>` and `-?>>` macros
 * Support key/value tables when destructuring
 * Add `match` macro for pattern matching
 * Add optional GNU readline support for repl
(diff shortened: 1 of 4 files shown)


thatismatt:
Just added `-?>>`. And I've noted these new macros in the changelog.

technomancy:
Could you add this to `reference.md` as well? Thanks. Looks good other than that.

thatismatt:
I've rebased this on the recent commits. Would someone be happy to merge it?

Now quote is in, should these (and other macros?) be rewritten to use quote?

technomancy:
I think it looks great. We can take a pass at converting the macros to using quote all at once after this is merged.

PR #129, opened 2018-12-27
Add a man page for Fennel

technomancy:
Use `man -l fennel.1` to read it.


technomancy:
Here's an HTML rendering: https://p.hagelb.org/fennel-manual.html

PR #130, opened 2018-12-29
Allow mangled globals in strict mode

thatismatt:
When inferring the allowed globals for the compilation environment we should include mangled names so that mangled identifiers don't get rejected as unknown globals.

This PR includes a regression test to illustrate the problem I want to fix.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -1739,7 +1739,10 @@ end
 
 local function currentGlobalNames(env)
     local names = {}
-    for k in pairs(env or _G) do table.insert(names, k) end
+    for k in pairs(env or _G) do
+       k = globalUnmangling(k)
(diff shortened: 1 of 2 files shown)


technomancy:
Good catch; thanks.

This seems like a good fix, but what I can't figure out is that the change in d39b3dd48a1b28fb76d0ab91511b73721087b4ac *should* have taken care of this for Lua 5.2+... but it didn't. I can't figure out why, but since that fix won't work for 5.1 anyway (`__pairs` metamethod is broken pre-5.2) I figure we need this.

thatismatt:
I think the reason is that the environment lookup happens after compilation to Lua, and it is that step that converts to the mangled form.

PR #132, opened 2019-01-01
automatically puts list around functions in threading macros if absent

noisesmith:
This replicates the familiar behavior of the threading macros with Clojure.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -2062,21 +2062,24 @@ end
 local stdmacros = [===[
 {"->" (fn [val ...]
         (var x val)
-        (each [_ elt (ipairs [...])]
-          (table.insert elt 2 x)
-          (set x elt))
+        (each [_ e (ipairs [...])]
+          (let [elt (if (list? e) e (list e))]
+            (table.insert elt 2 x)
+            (set x elt)))
         x)
  "->>" (fn [val ...]
          (var x val)
-         (each [_ elt (pairs [...])]
-           (table.insert elt x)
-           (set x elt))
+         (each [_ e (pairs [...])]
+           (let [elt (if (list? e) e (list e))]
+             (table.insert elt x)
+             (set x elt)))
          x)
  "-?>" (fn [val ...]
          (if (= 0 (# [...]))
              val
              (let [els [...]
-                   el (table.remove els 1)
+                   e (table.remove els 1)
+                   el (if (list? e) e (list e))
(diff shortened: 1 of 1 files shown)


technomancy:
I guess the only reason this wasn't included in the first pass was that while it's easy to add it in, it's impossible to remove later on if we decide it's not a good idea. Because of this, we should think a little more about whether it's the right thing for Fennel and not just do it because it can be done easily.

I don't feel strongly about it either way, but to me the advantage of the current way is that the code looks closer to its actual meaning. To people who haven't already used Clojure, it's easier to see the correspondence between the input and the resulting macroexpansion when the parens are there. The downside is ... slightly more typing, I guess.

In Clojure, you commonly see forms like `(-> m :key :word :inside)` because keywords are functions. In this case, the lack of parens emphasizes the "path to nested data" pattern which is the heart of what's really going on. In Fennel this pattern cannot be implemented with `->` because we don't have keyword functions, but we do have `(. m :key :inside :table)` syntax which accomplishes the same thing using a macro which is more specifically suited for this case. Because of this difference, IMO the rationale for making the parens optional is not as strong.

noisesmith:
If that's the prevailing view I'm OK with not merging this feature. We might want to look at the current behavior:

```
(cmd)>> (fn *2 [x] (* x 2))
#<function: 0x55a859f8af80>
(cmd)>> (-> 5 *2)
#<function: 0x55a859f8af80>
(ins)>> ((-> 5 *2))
[string "local _2a2 = ___replLocals___['_2a2']..."]:3: attempt to perform arithmetic on local 'x' (a nil value)
stack traceback:
  [string "local _2a2 = ___replLocals___['_2a2']..."]:3: in function ?
  /usr/local/share/lua/5.2/fennel.lua:1902: in function ?
  [C]: in function 'xpcall'
  /usr/local/share/lua/5.2/fennel.lua:1902: in function 'repl'
  /usr/local/bin/fennel:110: in main chunk
  [C]: in ?
```

if we don't want to coerce the forms to lists, perhaps we should explicitly error with an informative message when compiling if they aren't lists

technomancy:
> if we don't want to coerce the forms to lists, perhaps we should explicitly error with an informative message when compiling if they aren't lists

That's a great point; a better message is needed here if we keep the current behavior.

technomancy:
Talked about it on the list and it sounds like we should get this in. Thanks!

PR #135, opened 2019-01-12
Check for fishy redefinitions that might be bad

bakpakin:
It was possible in Fennel to create bindings that may be 'over-shadowed' by a macro or special form.
When used in a function call, a binding that shares a name with a special form cannot be used, as use
in the function position will be interpreted by the compiler as a special form or macro.

```
(global x 1)
(let [let print] (let [x let] x))
```
The above code is an example where the symbol let appears 4 times. The first let is a macro name, and the second introduces a binding named let. The third invocation of let will again be a macro invocation, even though a binding named let was made. The fourth let will be bound to print as expected.

Such code is confusing and surprising, so in this pull request, the compiler disallows all such code.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -928,6 +928,9 @@ local function destructure(to, from, ast, scope, parent, opts)
     -- Recursive auxiliary function
     local function destructure1(left, rightexprs, up1)
         if isSym(left) and left[1] ~= "nil" then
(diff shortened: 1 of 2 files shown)


technomancy:
Looks great; this could have cleared up the confusion about the attempted redefinition of `//` that we saw today in the channel.

PR #136, opened 2019-01-13
Update primer for post-release fennel

kenrestivo:
Since fennelview is now included, update docs to reflect that.


technomancy:
Nice; that is clearer.

PR #139, opened 2019-01-22
Addressing #139

brandonchartier:


technomancy:
Good catch; thanks! All that jumping back and forth between Fennel and Lua must have gotten me muddled.

PR #140, opened 2019-01-29
Allow macros to detect sequence literals

Benaiah:
This commit tags sequence literals (e.g., `[:a :b :c]`) with a specific metatable `SEQUENCE_MT`, and provides a `sequence?` function to macros to detect sequence literals. This allows macros which create data structures other than Lua tables from literals to see the source code's intent when deciding what kind of structure to create - even in the otherwise-thorny case of the empty table (`[]` vs. `{}`). My motivation for this change is working with Fengari - JS has distinct Object/Array types, so in order for a macro which uses a literal to create a JS structure, we have to be able to distinguish between the two kinds of literals, which without this change are not perfectly distinguishable.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -44,6 +44,7 @@ local LIST_MT = { 'LIST',
         return '(' .. table.concat(strs, ', ', 1, #self) .. ')'
     end
 }
+local SEQUENCE_MT = { 'SEQUENCE' }
 
 -- Load code with an environment in all recent Lua versions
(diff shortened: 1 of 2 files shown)


technomancy:
As long as this is strictly understood as a compile-time concern this seems pretty sensible.

Can you add a mention at the bottom of reference.md to mention the new predicate?

Benaiah:
I've updated this PR with the requested description of `sequence`?

PR #155, opened 2019-05-14
Destructuring shorthand for variables named the same as keys

Benaiah:
This change adds the following syntax:

```fennel
(let [tbl {:a-long-key-name :a-value}
      {: a-long-key-name} tbl]
  a-long-key-name) ;; => "a-value"
```

This is equivalent to the following code in the current version of Fennel:

```fennel
(let [tbl {:a-long-key-name :a-value}
      {:a-long-key-name a-long-key-name} tbl]
  a-long-key-name) ;; => "a-value"
```

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -1007,7 +1007,7 @@ local function destructure(to, from, ast, scope, parent, opts)
                     destructure1(left[k+1], {subexpr}, left)
                     return
                 else
-                    if type(k) == "string" and tostring(v) == ":" then v = sym(k) end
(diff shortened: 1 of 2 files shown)


bakpakin:
The implementation of this is impressivley short :), Looks good to me

Benaiah:
Don’t merge this or #156 yet - see https://github.com/bakpakin/Fennel/pull/156#issuecomment-492489388

Benaiah:
I've updated this PR to make the syntax match #156 - see https://github.com/bakpakin/Fennel/pull/156#issuecomment-492740927 for more details.

PR #156, opened 2019-05-15
Table literal shorthand for keys named the same as variables

Benaiah:
This change adds the following syntax:

```fennel
(let [a-long-key-name :a-value
      tbl {: a-long-key-name}]
  tbl.a-long-key-name) ;; => "a-value"
```

This is equivalent to the following code in the current version of Fennel:

```fennel
(let [a-long-key-name :a-value
      tbl {:a-long-key-name a-long-key-name}]
  tbl.a-long-key-name) ;; => "a-value"
```

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -304,6 +304,9 @@ local function parser(getbyte, filename)
                     end
                     val = {}
                     for i = 1, #last, 2 do
+                        if tostring(last[i]) == ":" and isSym(last[i + 1]) then
+                            last[i] = tostring(last[i + 1])
+                        end
                         val[last[i]] = last[i + 1]
                     end
                 end
--- a/test.lua
+++ b/test.lua
@@ -137,6 +137,8 @@ local cases = {
         ["(let [x 17] (. 17))"]=17,
         -- table lookup with literal
         ["(+ (. {:a 93 :b 4} :a) (. [1 2 3] 2))"]=95,
+        -- table lookup with literal using matching-key-and-variable shorthand
+        ["(let [k 5 t {: k}] t.k)"]=5,
         -- set works with multisyms
         ["(let [t {}] (set t.a :multi) (. t :a))"]="multi",
         -- set works on parent scopes


Benaiah:
Perhaps this should use the same syntax as the destructuring does - `(let [k :v t {:k :}] t.k) ;; => "v"` instead of the current implementation of `(let [k :v t {: k}] t.k)`, or the destructuring form should use this syntax.

technomancy:
I think we need to have a clearer definition of what `:` means. It seems to be something along the lines of "you figure it out, compiler" but if we can be consistent about it being "you figure out the key" vs "you figure out the value" it would help.

Benaiah:
@technomancy agreed, and with that in mind I think the most straightforward way to fix these PRs is to refactor #155 to look like this one already does (e.g. `(let [t {:k :v} {: k} t] k) ;; => "v"`), so the `:` would be interpreted in both cases as "the key with the same name as the following symbol". This is shorter and clearer than the other form (currently implemented in #155), which requires an extra colon and would be interpreted as "the symbol with the same name as the previous key", which is inconsistent with this PR. If we were to refactor this PR to match #156 (e.g. `(let [k :v t {:k :}] t.k) ;; => "v"`), it would require extra characters and would be less clear IMO. `(local {:k :} t) k` seems confusing since the `k` symbol's first appearance is when the value is used, and you have to be very familiar with the syntax to follow that `k` is declared by the use of `:`. (`(local {: k} t)` is clearer in my mind - it's much clearer that a local variable is being declared with a "missing" key name.

Benaiah:
I've updated #155 to use the syntax implemented in this PR.

technomancy:
Yeah, I like the way you landed in the end here because making the key inferred means that the symbol is explicit, and that is very helpful for visually scanning the file to see where a given local came from. It works better with things like `idle-highlight-mode` in Emacs.

Nice work!

I'll get this merged along with its accompanying PR. Do you think you could do a follow-up where you update `reference.md`?

PR #158, opened 2019-05-17
Require whitespace before opening delimiter

Benaiah:
Require whitespace before opening delimiters - disallows both the following cases:

```lisp
(local x #(+ 1 $1))
(local y #(2))
(x(y)) ; used to return 3, now fails
(x (y)) ; still returns 3

(local z #[1 2 ((or unpack table.unpack) $1)])
(z[3 4]) ; used to return [1 2 3 4], now fails
(z [3 4]) ; still returns [1 2 3 4]
```

The other commit just adds the closing delimiter to the "uhexpected closing delimiter" error message.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -237,6 +237,7 @@ local function parser(getbyte, filename)
 
         -- Dispatch when we complete a value
         local done, retval
+        local whitespaceSinceDispatch = true
         local function dispatch(v)
             if #stack == 0 then
                 retval = v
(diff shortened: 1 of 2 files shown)


bakpakin:
Looks ok to me. Minor nitpick in the example, I believe it should be `(local y #2)`.

Also, while not strictly necessary, disallowing symbols before forms without whitespace allows us to
to add new prefix forms or reader macros without being backwards incompatible, so i'm totally for it.

Benaiah:
You're right - I've updated the description so the examples make sense to any future readers of the PR.

PR #160, opened 2019-05-18
Pattern match guards

technomancy:
This allows pattern match clauses to include guard expressions which will prevent them from matching when they evaluate to false/nil.

Example:
```lisp
(match [91 12 53]
  ([a b c] ? (= 5 a)) :will-not-match
  ([a b c] ? (= 0 (math.fmod (+ a b c) 2)) (= 91 a)) c) ; -> 53
```

The pattern consists of parens wrapping a regular pattern, followed by a `?` symbol followed by one or more guard expressions. This seems to conflict with the use of parens for multi-valued patterns, but it should be distinct for two reasons: the `?` symbol, but also the fact that the guard expressions will also have parens in them, and multi-valued patterns can never have parens inside parens.

I believe this makes the pattern matching system a fair bit more useful. I'm not completely sold on the use of the `?` symbol; I had `:?` before but I think the symbol is better? Open to suggestions for alternatives here.

guards guards

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,7 @@
 
 ## 0.3.0 / ??
 
+* Add support for guard clauses with `:?` in pattern matches.
 * Support completion in repl when `readline.lua` is available.
 * Add `--globals` and `--globals-only` options to launcher script.
 * Remove `luaexpr` and `luastatement` for a single `lua` special.
--- a/changelog.md
+++ b/changelog.md
@@ -2,7 +2,7 @@
 
 ## 0.3.0 / ??
 
-* Add support for guard clauses with `:?` in pattern matches.
+* Add support for guard clauses with `?` in pattern matches.
 * Support completion in repl when `readline.lua` is available.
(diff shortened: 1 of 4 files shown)


bakpakin:
The general syntax seems good to me. Racket has similar syntax but uses `:when` in place of the ? The question mark is terser but then :when reads very naturally IMO.
```clojure
(match [91 12 53]
  ([a b c] :when (= 5 a)) :will-not-match
  ([a b c] :when (= 0 (math.fmod (+ a b c) 2)) (= 91 a)) c) ; -> 53
```
But yeah, this looks at least as good as the alternatives I found for other lisps

technomancy:
Yeah, Racket's is nice, but I don't really like `:when` because the `when` macro has an implicit `do` in it, indicating it's for side-effects, while there's no such implication here.

bakpakin:
No strong feelings for `:when`; I say merge it.

technomancy:
Cool; thanks for taking a look.

Issue #161, opened 2019-05-29
match macro does not match nil correctly.


Benaiah:
The `match` macro appears to work inconsistently when matching `nil`. See the following example:

```lisp
(match nil
  _ ;; this should match
  "wildcard-match"

  (b ? (= b nil)) ;; EDIT - this maybe shouldn't match - see below
  "guard-match"

  nil ;; this should also match, but it's the only one that actually does
  "nil-match")
```

On the current master version of Fennel, this will always evaluate to `"nil-match"`, despite the fact that both the preceding cases should also match. (On the current _release_ of Fennel, which is slightly older, this actually causes the following compilation error: `Compile error in '=' unknown:?: expected at least two arguments`).

EDIT:

Turns out the second case (`"guard-match"`) is actually intended, and you _can_ match `nil` in a guard clause with the following construct:

```lisp
(match nil
  (?b ? (= ?b nil)) ;; this matches
  "guard-match")
```

The leading `?` in the binding symbol allows the macro to bind it to `nil`.

PR #162, opened 2019-05-30
Make _ match nil in the match macro

Benaiah:
Fixes part of #161. The other case mentioned in that issue is potentially a wontfix, as an existing way to use guard clauses with `nil` already exists.


technomancy:
Good catch! Can you change it so we use `when` to make it clearer that the `tset` call is a side-effect? Also the use of `is-*` comes from non-lisps; it would be more idiomatic to use `wildcard?` instead.

technomancy:
Perfect; thanks!

PR #163, opened 2019-06-01
Make multisyms work correctly in match macro

Benaiah:
Previously, the following code would fail due to a compilation error of `did not expect multisym`, instead of returning true:

```lisp
(let [x {:y :z}] (match :z x.y true _ false))
```

Now, the macro detects multisyms, checks if `x` is in scope (not `x.y`, which doesn't make sense to be in scope), and when possible adds a check that the value of the multisym pattern equals the value it's being matched against.

Additionally, `isMultiSym`/`multi-sym?` has been extended to work on AST symbols as well as strings.

Finally, a regression test has been added to `test.lua` that ensures the bug does not reappear.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -480,6 +480,9 @@ end
 -- from normal symbols is that they cannot be declared local, and
 -- they may have side effects on invocation (metatables)
 local function isMultiSym(str)
+    if isSym(str) then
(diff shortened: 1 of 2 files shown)


technomancy:
Nice catch; thanks! Great work with the fix.

PR #166, opened 2019-06-05
Add oneline option to fennelview

Benaiah:
Allows the use of `(fennelview X {:oneline true})`, which restricts fennelview's output to a single line and removes extra spaces around delimiters.

### DIFF
--- a/fennelview.fnl
+++ b/fennelview.fnl
@@ -146,11 +146,19 @@
 
 
 
+(fn one-line [str]
+  (-> str
+      (: :gsub "\n" " ")
+      (: :gsub "%[ " "[") (: :gsub " %]" "]")
+      (: :gsub "%{ " "{") (: :gsub " %}" "}")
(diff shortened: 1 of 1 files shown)


technomancy:
@bakpakin how do you feel about using `oneline` rather than `one-line` or `oneLine`? Lua convention seems to be to smush words together, but in other places we use camel case.

bakpakin:
Most of the camelcase is in Lua facing code I believe. Most of the camel case is in the Fennel module, which may have to bridge with Lua. I think we should try to stick to kebab case here, and maybe recommend that most fennel should use kebab case. We are kind of all over the place with naming conventions though.

Benaiah:
@technomancy @bakpakin I've updated the PR to use `one-line` instead. Thinking about this further, perhaps it should even be `one-line?` rather than `one-line` - what do you think?

bakpakin:
I thought a question mark suffix was for predicate functions. I think `:one-line` is fine.

PR #168, opened 2019-06-11
Include special form

bakpakin:
This pull request is for the `include` special form. The include special form reads a Fennel or Lua source file and splices it into the current Fennel compilation unit, usually either an output file a string that will be immediately compiled into a thunk. The semantics are similar to what was discussed on IRC, except that there is no pollution of lexical scope. This means that when used at the top level, `include` behaves almost exactly the same way as require from the user's point of view. `include` is also memoized for the current scope, so it can be used as a way of requiring dependencies.

For example, 

```clojure
(local fv (include "fennelview"))
(fv.fennelview [1 2 3 4])
```

Will embed the compiled Lua code of fennelview.fnl into the output. This can be used to generate portable scripts that are a single, self contained Lua source file.

`include` can also embed single file Lua dependencies perfectly into the generated source by wrapping their source in a function expression. `include` does not rewrite invocations of `require`, so any module loaded via `require` must be available at runtime.

## Caveats

The source file included by `include` can "see" symbols in the current scope where it was included. I don't believe that this is much of a problem, especially if `include` is used primarily at the beginning of files.

`include` can also be used to include multiple copies of the same code, as the memoization occurs only per scope. For example, the following will include the source code from `a.fnl` only once.

```clojure
(include "a.fnl")
(include "a.fnl")
(include "a.fnl")
```

But we can also include three copies of `a.fnl` by putting each `include` inside a do.

```clojure
(do (include "a.fnl"))
(do (include "a.fnl"))
(do (include "a.fnl"))
```

### DIFF
--- a/fennel
+++ b/fennel
@@ -19,7 +19,7 @@ Run fennel, a lisp programming language for the Lua runtime.
   --add-fennel-path  PATH : Add PATH to fennel.path for finding Fennel modules
   --globals G1[,G2...]    : Allow these globals in addition to standard ones
   --globals-only G1[,G2]  : Same as above, but exclude standard ones
-  --include               : Inline required modules in the output
+  --require-as-include    : Inline required modules in the output
 
   --eval SOURCE (-e)      : Evaluate source code and print the result
 
@@ -82,8 +82,8 @@ for i=#arg, 1, -1 do
     elseif arg[i] == "--globals-only" then
         allowGlobals(table.remove(arg, i+1))
         table.remove(arg, i)
-    elseif arg[i] == "--include" then
-        options.include = true
+    elseif arg[i] == "--require-as-include" then
+        options.requireAsInclude = true
         table.remove(arg, i)
     end
 end
--- a/fennel
+++ b/fennel
@@ -19,6 +19,7 @@ Run fennel, a lisp programming language for the Lua runtime.
   --add-fennel-path  PATH : Add PATH to fennel.path for finding Fennel modules
   --globals G1[,G2...]    : Allow these globals in addition to standard ones
   --globals-only G1[,G2]  : Same as above, but exclude standard ones
+  --include               : Inline required modules in the output
 
   --eval SOURCE (-e)      : Evaluate source code and print the result
 
@@ -80,6 +81,9 @@ for i=#arg, 1, -1 do
         table.remove(arg, i)
     elseif arg[i] == "--globals-only" then
         allowGlobals(table.remove(arg, i+1))
+        table.remove(arg, i)
+    elseif arg[i] == "--include" then
+        options.include = true
         table.remove(arg, i)
     end
 end
--- a/fennel.lua
+++ b/fennel.lua
@@ -2250,7 +2250,7 @@ SPECIALS['include'] = function(ast, scope, parent)
             for _, val in p do table.insert(forms, val) end
             e = compile1(forms, scope, parent, {nval = 1})[1]
         else
-            e = expr('(function()' .. s .. ' end)()')
+            e = expr('(function()' .. s .. ' end)()', 'expression')
         end
 
         local memoize = once(e, ast, scope, parent)
--- a/fennel.lua
+++ b/fennel.lua
@@ -404,6 +404,11 @@ end
 -- Compilation
 --
 
+-- Top level compilation bindings.
+local rootChunk
+local rootScope
+local rootOptions
+
 -- Create a new Scope, optionally under a parent scope. Scopes are compile time constructs
 -- that are responsible for keeping track of local variables, name mangling, and macros.
 -- They are accessible to user code via the '*compiler' special form (may change). They
@@ -829,7 +834,6 @@ end
 local function compile1(ast, scope, parent, opts)
     opts = opts or {}
     local exprs = {}
-    scope.chunk = parent
 
     -- Compile the form
     if isList(ast) then
@@ -936,7 +940,6 @@ local function compile1(ast, scope, parent, opts)
         assertCompile(false, 'could not compile value of type ' .. type(ast), ast)
     end
     exprs.returned = true
-    scope.chunk = nil
     return exprs
 end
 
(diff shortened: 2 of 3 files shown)


Benaiah:
> The source file included by include can "see" symbols in the current scope where it was included. I don't believe that this is much of a problem, especially if include is used primarily at the beginning of files.

This could be solved by using `load` with an explicitly set environment, couldn't it? Though I suppose macros would still leak in since they're run at compile time (this can be solved by having macros be required by name, as I've suggested before).

As I mentioned on IRC, I'm pretty hesitant about textual inclusion without it being quite well sealed off from the surrounding environment. The ability of `include` to bind an arbitrary number of unknown symbols in the current file's scope from another file seems like a massive antipattern, and requiring `do` be wrapped around it to prevent that means that new users who don't fully understand the implications will be using the more dangerous version by default.

Ideally, Fennel code in one file would never break because you accidentally shadowed a symbol in another file, unless you're knowingly using globals. The only thing that currently breaks that assumption is the macro system, which can be changed to require importing the macros into specific symbols, instead of "whatever they're named in the macro file". If `include` is merged as-is, it will completely and irrevocably destroy that feature of the language.

I think compile-time including is a very strong feature, and it wouldn't take drastic changes to do so in a way that doesn't have the caveats I've mentioned. I suspect that, if this is merged, it will likely become a widely-used Fennel feature (because single-file distribution is so handy in most Lua environments), and the `include` special form as written in this PR encourages the style that is most likely to cause bugs and "spooky action at a distance". It encourages it because `(include ...)` is shorter and more obvious than `(do (include ...))` - though `(do (include ...))` is _still_ not 100% safe.

We could add something that works like this, but naming it `include` does not seem to adequately warn of its caveats. If it was named something like `dangerously-include` or `raw-include` I would feel much more comfortable merging it.

technomancy:
> This means that when used at the top level, `include` behaves almost
> exactly the same way as require from the user's point of view

Would it be possible for this to function as a compile-time switch which
instead of introducing a new special form could turn `require` into the
text-including special form?

That might help avoid the concern Benaiah raises about people writing
code that allows the included file to pollute the scope. During
development, you would use `require` as a normal file with the standard
module system and scoping rules, then when ready to cut a release, you
would cut a release using the --require-include flag or whatever it ends
up being called.

(A nice bonus of this is it might work recursively; a module which
requires another module should be able to include it without making
changes to the source.)

> The source file included by `include` can "see" symbols in the current
> scope where it was included.

If strict global checking is turned on, shouldn't this be impossible for
included Fennel modules? The compiler will see a reference to an unknown
identifier and it will error out rather than letting it leak thru. It
also wouldn't be a problem with well-written Lua files; if they have
`local foo` it would shadow an existing `foo` binding. It would be a
problem with bad Lua files in that it would turn locals in the
includer's scope into what amounts to globals for the includee's scope,
but such Lua modules are already fairly problematic.

bakpakin:
> If strict global checking is turned on, shouldn't this be impossible for
included Fennel modules? 

Yes, the problem is with poorly written Lua as you mentioned. I don't think it is a big issue, but I thought I would mention it.

I think some of the caveats above can be worked out as well, such as the double includes and minor scoping issues. I will probably push a few updates to the current PR as well that I hope will eliminate these.

> Would it be possible for this to function as a compile-time switch which instead of introducing a new special form could turn `require` into the text-including special form?`

Yes, it would be possible and easy. My hesitation was that by making `require` a special form, it would make it harder to do a runtime require. A separate special form was an easy way to make sure everything was backwards compatible. If we add an option to overload require with a special form, I think it would be useful to also pass in a list of modules to the compiler that would be loaded at runtime (use regular require). Like a `--runtime-modules` flag which takes a comma separated list of modules to not include into the output source.

Benaiah:
Yeah, I agree that we should have an option to list modules to require at runtime. We'll also need to exclude any `require` forms that aren't at top level with literal arguments, and perhaps error or warn when we do so. Fengari's webpack loader [does something quite similar](https://github.com/fengari-lua/fengari-loader#how-does-it-work) (though AFAIK it doesn't have a way to turn the error off, which is annoying).

Benaiah:
@bakpakin is there anything preventing including both the `(include ...)` special and the `--include` CLI option? The latest change removes the `include` special (unless I've misunderstood the code). The new `--include` behavior is great, but I think the original `(include ...)` special is useful as well for cases where you want to be explicit about which files are being loaded at run time vs. compile time.

Regarding the option name - if I didn't know better, `--include` would read to me as an option that would take a path argument and include that file before the rest of the compiled file. I'm not sure how useful that would be in Fennel, so I'm not proposing adding that feature right now, but I think the CLI option would be clearer if it was named something like `--include-required` or `--require-as-include`. This point feels like it could be bikeshedding, so I'm not strongly opposed to leaving the option name as is.

Currently, the special used by the `--include` option simply calls `tostring` on the argument to `require`. This breaks in strange ways if the argument is not a string literal:

- the expression `(let [fv :fennelview] (require fv))` with `--include` on results in a compiler error because the module `fv` was not found, unless an `fv` package exists
- if an `fv` package _does_ exist, it will be included instead of including `:fennelview`, raising an error, or leaving the require dynamic
- if the argument to `require` is not a string literal or a symbol, the compiler attempting to include the `tostring`ed expression output will almost certainly result in a compiler error due to no such module being found (e.g., `(require (.. :f :v))` breaks with `could not find module (.., f, v)`)
- the previous problem occurs even if the form can be evaluated entirely at compile time (e.g., `(macro fv! [] :fennelview) (require (fv!))` causes the compiler error `could not find module (fv!)`)

To fix these issues, the argument to `require` needs to be evaluated as fully as can be done at compile time, and the resulting arguments should then be checked to see if they are static values (i.e., the compiler shouldn't include a file by matching the _symbol_ name, even if such a file exists). If they're not, the compiler should raise a descriptive error or leave the require dynamic (this should be configurable).

bakpakin:
Yeah, all of these suggestions are pretty good. I'll add back `include` and change the name of `--include` to `--require-as-include`. Also will try to make evaluation a bit smarter and optionally fallback to normal require if needed (`include` will not have a fallback).

There are also two kinds of include failures to consider. The first is when `include` cannot deduce statically what module string to include, as given above. The second is when include cannot find the module given. It's not clear to me if `--require-as-include` should provide a uniform fallback for both these errors, or be configurable.

PR #172, opened 2019-07-12
Pass options when `fennel somefile.fnl` is executed

jaawerth:
Ensures any flags that affect compilation, etc are respected.


technomancy:
Good catch; thanks!

PR #173, opened 2019-07-12
Add metadata and docstrings

### LINKED ISSUE #154: docstrings
Would it be possible to add support for proper docstrings? Or even cooler - metadata in general.

jaawerth:
resolves #154 

Implemented function metadata and docstring capability! The metadata table (which has weak keys, so it won't stop functions from being GC'd) can be extended further to store whatever you want in association with a function.

## Example
```clojure
Fennel 0.3.0 - Welcome to Fennel!
; regular function
>> (fn identity [x] "==> x" x)
#<function: 0x7feceff401e0>
>> (doc identity)
(identity x)
  ==> x

; higher-order function
>> (fn always [...]
.. "return a function that always returns everything passed in"
.. (let [args [...]] (fn [] "constant returner" (table.unpack args))))
#<function: 0x7fecf1027110>
>> (doc always)
(always ...)
  return a function that always returns everything passed in
>> (doc (always 1 2 3 4 5))
(_0_)
    constant returner

>> ((always 1 2 3 4 5))
1       2       3       4       5

; lambdas work, but the magic syntax does weird things to the generated invocation example
>> (λ greet [?greeting ?name]
.. "Say hi!"
.. (string.format "%s, %s" (or greeting :Hello) (or name :Stranger)))
#<function: 0x7fecefd5c830>
>> (greet)
"Hello, Stranger"
>> (doc greet)
(greet _3fgreeting _3fname)
  Say hi!
```
### new flags/options
- `--metadata` and `--no-metadata` on the CLI
- `useMetadata` in the compiler API (defaults to false)
- REPL defaults `useMetadata` to true

### Metadata
Introduces table on `fennel.metadata` with weakly referenced keys. When metadata is enabled by passing `useMetadata = true` to the options (defaults to true in REPL), or passing the `--metadata` flag from the CLI, function and lambda declarations will automatically store this.

**NOTE:** This also means, when metadata enabled, that function/lambda declarations will automatically require `fennel`.

### `fennel.doc`

Introduces a new API function `fennel.doc` which will print a function's metadata (or `nil` if there is none, or if the function doesn't exist). This is set as a local in the REPL for convenience.

### When compiling or running fennel
Docstrings are parsed out, but not set on metadata, because there is a performance hit to creating an entry in a weak metadata table for every function as well as the memory cost of storing it. As such, docstrings default to a NOOP.

### In the REPL
Metadata is enabled automatically, so values are stored in the metadata.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -45,19 +45,6 @@ local LIST_MT = { 'LIST',
     end
 }
 local SEQUENCE_MT = { 'SEQUENCE' }
-local METADATA_MT = {
-  __mode = 'k',
-  __index = {
-    get = function(self, tgt, key)
-      if self[tgt] then return self[tgt][key] end
-      return nil
-    end,
-    set = function(self, tgt, key, value)
-      self[tgt] = self[tgt] or {}
-      self[tgt][key] = value
-      return tgt
-    end
-  }}
 
 -- Load code with an environment in all recent Lua versions
 local function loadCode(code, environment, filename)
@@ -768,7 +755,18 @@ end
 -- module-wide state for metadata
 -- create metadata table with weakly-referenced keys
 local function makeMetadata()
-  return setmetatable({}, METADATA_MT)
+    return setmetatable({}, {
+        __mode = 'k',
+        __index = {
+            get = function(self, tgt, key)
+                if self[tgt] then return self[tgt][key] end
+            end,
+            set = function(self, tgt, key, value)
+                self[tgt] = self[tgt] or {}
+                self[tgt][key] = value
+                return tgt
+            end
+        }})
 end
 
 local metadata = makeMetadata()
@@ -2246,7 +2244,6 @@ module.makeSearcher = function(options)
 end
 
 -- Add metadata and docstrings to fennel module
-module.makeMetadata = makeMetadata
 module.metadata = metadata
 module.doc = doc
 
--- a/fennel.lua
+++ b/fennel.lua
@@ -45,6 +45,19 @@ local LIST_MT = { 'LIST',
     end
 }
 local SEQUENCE_MT = { 'SEQUENCE' }
+local METADATA_MT = {
+  __mode = 'k',
+  __index = {
+    get = function(self, tgt, key)
+      if self[tgt] then return self[tgt][key] end
+      return nil
+    end,
+    set = function(self, tgt, key, value)
+      self[tgt] = self[tgt] or {}
+      self[tgt][key] = value
+      return tgt
+    end
+  }}
 
 -- Load code with an environment in all recent Lua versions
 local function loadCode(code, environment, filename)
@@ -752,6 +765,16 @@ local function flatten(chunk, options)
     end
 end
 
+-- module-wide state for metadata
+-- create metadata table with weakly-referenced keys
+local function makeMetadata()
+  return setmetatable({}, METADATA_MT)
+end
+
+local metadata = makeMetadata()
+local doc = function(tgt)
+  print(metadata:get(tgt, 'docstring'))
+end
 -- Convert expressions to Lua string
 local function exprs1(exprs)
     local t = {}
@@ -1182,12 +1205,13 @@ SPECIALS['values'] = values
 -- The fn special declares a function. Syntax is similar to other lisps;
 -- (fn optional-name [arg ...] (body))
 -- Further decoration such as docstrings, meta info, and multibody functions a possibility.
-SPECIALS['fn'] = function(ast, scope, parent)
+SPECIALS['fn'] = function(ast, scope, parent, options)
     local fScope = makeScope(scope)
     local fChunk = {}
     local index = 2
     local fnName = isSym(ast[index])
     local isLocalFn
+    local docstring
     fScope.vararg = false
     if fnName and fnName[1] ~= 'nil' then
         isLocalFn = not isMultiSym(fnName[1])
@@ -1221,10 +1245,23 @@ SPECIALS['fn'] = function(ast, scope, parent)
             assertCompile(false, 'expected symbol for function parameter', ast)
         end
     end
+    if type(ast[index + 1]) == 'string' and index + 1 < #ast then
+      index = index + 1
+      docstring = ast[index]
+      if rootOptions.useMetadata then
+        docstring = docstring -- account for newlines and nested [[ ]] multiline strings
+                :gsub('\n', '\\n')
+                :gsub('[[](=*)[[]', '[=%1['):gsub('[]](=*)[]]', ']=%1]')
+        -- splice in argslist and format docstring contents
+        local argSep = #argNameList > 0 and ' ' or ''
+        local invokeStr = string.format("(%s%s%s)", fnName, argSep, table.concat(argNameList, ' '))
+        docstring = string.format("%s\n  %s", invokeStr, docstring)
+      end
+    end
     for i = index + 1, #ast do
         compile1(ast[i], fScope, fChunk, {
             tail = i == #ast,
-            nval = i ~= #ast and 0 or nil
+            nval = i ~= #ast and 0 or nil,
         })
     end
     if isLocalFn then
@@ -1234,8 +1271,16 @@ SPECIALS['fn'] = function(ast, scope, parent)
         emit(parent, ('%s = function(%s)')
                  :format(fnName, table.concat(argNameList, ', ')), ast)
     end
+
     emit(parent, fChunk, ast)
     emit(parent, 'end', ast)
+
+    if rootOptions.useMetadata and docstring then
+      local docSetter = string.format('require("fennel").metadata:set(%s, "docstring", [[%s]])',
+        fnName, docstring)
+      emit(parent, docSetter)
+    end
+
     return expr(fnName, 'sym')
 end
 
@@ -1849,7 +1894,7 @@ local function compileStream(strm, options)
     rootOptions = options
     for i = 1, #vals do
         local exprs = compile1(vals[i], scope, chunk, {
-            tail = i == #vals
+            tail = i == #vals,
         })
         keepSideEffects(exprs, chunk, nil, vals[i])
     end
@@ -1997,6 +2042,10 @@ local function repl(options)
         opts.allowedGlobals = currentGlobalNames(opts.env)
     end
 
+    if opts.useMetadata == nil then
+      opts.useMetadata = true
+    end
+
     local env = opts.env and wrapEnv(opts.env) or setmetatable({}, {
         __index = _ENV or _G
     })
@@ -2044,7 +2093,8 @@ local function repl(options)
         -- we do some source munging in order to save off locals from each chunk
         -- and reintroduce them to the beginning of the next chunk, allowing
         -- locals to work in the repl the way you'd expect them to.
-        env.___replLocals___ = env.___replLocals___ or {}
+        local replDoc = opts.useMetadata and require("fennel").doc or nil
+        env.___replLocals___ = env.___replLocals___ or { doc = replDoc }
         local splicedSource = {}
         for line in luaSource:gmatch("([^\n]+)\n?") do
             table.insert(splicedSource, line)
@@ -2114,6 +2164,7 @@ local function repl(options)
                 sourcemap = opts.sourcemap,
                 source = srcstring,
                 scope = scope,
+                useMetadata = opts.useMetadata,
             })
             if not compileOk then
                 clearstream()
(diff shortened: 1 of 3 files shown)


jaawerth:
... so I was running `make` when testing this instead of explicitly running `make testall` and missed the one that failed in the CI. Whoops! I'll address that tomorrow and then this should hopefully be ready for review.

jaawerth:
Fixed! Just had to tweak part of the new test section to use `unpack` or `table.unpack` based on availability.

fennel.lua
```diff
@@ -752,6 +765,16 @@ local function flatten(chunk, options)
     end
 end
 
+-- module-wide state for metadata
+-- create metadata table with weakly-referenced keys
+local function makeMetadata()
+  return setmetatable({}, METADATA_MT)
```
technomancy: Could we make `METADATA_MT` local to this function? It looks like it isn't used anywhere else.

fennel.lua
```diff
@@ -1221,10 +1245,23 @@ SPECIALS['fn'] = function(ast, scope, parent)
             assertCompile(false, 'expected symbol for function parameter', ast)
         end
     end
+    if type(ast[index + 1]) == 'string' and index + 1 < #ast then
+      index = index + 1
+      docstring = ast[index]
+      if rootOptions.useMetadata then
+        docstring = docstring -- account for newlines and nested [[ ]] multiline strings
+                :gsub('\n', '\\n')
+                :gsub('[[](=*)[[]', '[=%1['):gsub('[]](=*)[]]', ']=%1]')
```
technomancy: I think the `[[this style]]` of strings is valid for Lua but multiline Fennel strings just use normal quotes?

fennel.lua
```diff
@@ -1221,10 +1245,23 @@ SPECIALS['fn'] = function(ast, scope, parent)
...
+        -- splice in argslist and format docstring contents
+        local argSep = #argNameList > 0 and ' ' or ''
+        local invokeStr = string.format("(%s%s%s)", fnName, argSep, table.concat(argNameList, ' '))
```
technomancy: What if the docstrings and the arg names were stored in different metadata slots? That way when you did doc lookup on a function, you could get argument names even on functions that didn't have docstrings.

test.lua
```diff
@@ -584,6 +584,40 @@ for k, v in pairs(quoting_tests) do
     end
 end
 
+---- docstring tests ----
+
+print("Running tests for docstrings...")
+local docstring_tests = {
+  ['(doc (fn [] :A 1))'] = {'(_0_)\\n  A', 'anonymous fn docstring mistmatch'},
+  ['(doc (λ [a] :B :ret-str))'] = {'(_0_ a)\\n  B', 'anonymous lambda docstring mistmatch'},
```
technomancy: I'm not sure how I feel about `_0_` showing up here... On the other hand, leaving it out seems weird, and I'm not sure what it would make sense to replace it with... maybe the `tostring` of the function?

`(#<function: 0x5623043dd020> a)\n\n   B`

technomancy:
> lambdas work, but the magic syntax does weird things to the generated invocation example

This isn't actually a `lambda` problem; it has to do with any mangled identifiers:

```lisp
>> (fn abc [def-xyz] "returns fifty nine" 59)
#<function: 0x5623043e1b50>
>> (doc abc)
(abc def_xyz)
  returns fifty nine
```

So we ought to see if we can unmangle the arg names before storing them in metadata.

fennel.lua
```diff
@@ -2194,6 +2245,11 @@ module.makeSearcher = function(options)
    end
 end
 
+-- Add metadata and docstrings to fennel module
+module.makeMetadata = makeMetadata
```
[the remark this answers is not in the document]
jaawerth: When I first put it together, I had some thought about making it possible for folks to have the metadata live elsewhere. That's probably overcomplicating it, though, particularly since I didn't provide a means of changing where it's automatically stored in function definitions.

I'll remove it now - we can always make it more configurable in future if we find a good usecase for that.

fennel.lua
jaawerth: Good call - I was a bit undecided about doing that right out of the gate but it absolutely makes more sense to do it that way. Plus, having those other items in the metadata will probably open up other use-cases.

test.lua
jaawerth: I had the same thought, and figured a subsequent PR could clean that up a bit but in the meantime it was better than not having it at all.

My initial thought was replacing `_n_` with something like `<anonymous>` so it would look like `(<anonymous> a)\n  B` but I can see the benefit of having the address in there.

That said, I feel like if we're going to explicitly handle that case we should make it clear to the end-user that it's showing up that way because there wasn't a function name. Thoughts?

technomancy: Yeah, explicitly indicating it's anonymous is good. What about `#<anonymous>` since the `#<>` notation is already used for "unreadable" things? But we could leave it as-is for now and come back to it later; that's fine.

fennel.lua
jaawerth: That's there to account for when emitting the lua that sets the docstring on line 1279
```lua
local docSetter = string.format('require("fennel").metadata:set(%s, "docstring", [[%s]])',
```

since, if someone were to put `[[ ]]` in the docstring without the above escaping of `[[ ]]`, it'd break. Though, upon reflection, `:gsub('\n', '\\n')` may be unnecessary. I'll add some tests for multiline docstrings and see. I'll also throw in some tests for docstrings with `[[ ]]` just to futureproof it.

I suppose I could just pass the docstring into `compile` or `compile1` to just let the compiler's existing string-juggling handle all that?

jaawerth: Actually, leaving in `:gsub('\n', '\\n')`, removing the `[[ ]]` handling and changing line 1279 to
```lua
local docSetter = string.format('require("fennel").metadata:set(%s, "docstring", "%s")',
```
works! It makes the docstring less readable in the compiled lua than using `[[ ]]` but also means emitting fewer lines of code when there are multiline docstrings so it's kind of an even trade.

technomancy: Oh, I see. I misunderstood the purpose of that snippet, so go ahead and ignore my comment.

Benaiah: Agreed - unless I’m mistaken, `[[strings like these]]` aren’t currently valid in Fennel, and it’s a syntax that collides with Fennel’s sequential-table literals.

test.lua
jaawerth: `#<anonymous>` sounds good to me! This will be easy enough to throw into the `doc` function when I update the branch to store `fnName` and `argsList` as separate metadata fields, since the work of assembling the pretty output will be moved to the `doc` function - I can just do something like
```lua
fennel.doc = function doc(fn)
  local docstring = metadata:get(fn, 'docstring') or '<undocumented>'
  local fnName = metadata:get(fn, 'fnName') or '#<anonymous>'
  local argsList = metadata:get(fn, 'argsList') or {}
  local invokeStr = string.format('(%s%s%s)', fnName, #argsList > 0 and ' ' or '', table.concat(argsList, ' '))
  print(string.format('%s\n  %s', invokeStr, docstring))
end
```

jaawerth:
I might be getting ahead of myself, but as I started finalizing this and getting excited about all the cool things we'll be able to do with the metadata (like `pre` and `post`, maybe even some basic `core.spec.alpha` stuff for fennel?) it occurred to me that, since a lot of that stuff should live in userland, we might want to prefix/namespace all the built-in metadata fields with something that, by convention, devs shouldn't use, just to get ahead of any future conflicts with native fields we may want to add down the road.

Maybe something like `:fnl/`, so we have `:fnl/docstring`, `:fnl/args-list`, etc?

I'm going to plan on throwing this into my final update but feedback welcome if anyone has any concerns/objects/alternate convention formats.

technomancy:
I don't think it's necessarily needed to namespace built-in stuff like docstrings, but it might be smart to use namespacing anyway in order to encourage others to namespace their keys once they start writing third-party metadata keys. Leading by example, or something. The `:fnl/` prefix has my thumbs up.

fennel.lua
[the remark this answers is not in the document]
jaawerth: Removed!

test.lua
jaawerth: I tried implementing this for the update I just pushed, but it turned out to be a bit extra confusing due to the realization that functions set to local/global vars or table properties will be treated as anonymous:

E.g.
```clojure
(global foo (fn [] :stuff) (doc foo) ; -> (#<anonymous)\n  <nil>
(local tbl {:foo (fn [] "do things!" :things)})
(doc tbl.foo) ; -> #(anonymous)\n  do things!
```

Detecting those kinds of assignments seems like a nontrivial addition so I'm going to leave it off for now.

jaawerth:
Update is ready for perusal. Note that to avoid confusion I haven't rebased off of the new master yet but once we're happy with this round of changes I can do so!

fennel.lua
jaawerth: Done!

technomancy:
Nice work. I think punting on the anonymous function bits for a future patch makes sense. That might get tricky and need reaching into the `debug` table black magic but it can happen at a future time.

One nit; can we add tests for mangled names? Once we get that in, I'd say we're good to merge this.

jaawerth:
Done! Barring any further feedback or rebasing off current master it should be good to go.

technomancy:
Oh, sorry--what I meant was that we should store the unmangled names in the arglist and docstring metadata as these are closer to what was intended.

jaawerth:
Ohh gotcha. The "add tests" parts threw me off since the unmangled names isn't implemented yet, but that should be fairly quick to implement. I'm off camping for a few days tomorrow, so if I don't have time in the morning I'll knock it out on Monday.

test.lua
```diff
@@ -584,6 +584,59 @@ for k, v in pairs(quoting_tests) do
...
---- docstring tests ----
+
+print("Running tests for metadata and docstrings...")
+local docstring_tests = {
+    ['(doc (fn [] :A 1))'] = {'(_0_)\\n  A', 'no anonymous fn docstring mistmatch'},
+    ['(doc (λ [a] :B :ret-str))'] = {'(_0_ a)\\n  B', 'no anonymous lambda docstring mistmatch'},
+    -- TODO: set correct fnName based on assigned var/prop name and use following test
+    -- ['(local foo (fn [a] :C 1)) (doc foo)'] = {'(foo a)\\n  C', 'named/non-local fn docstring mismatch'},
+    ['(doc (fn foo [a] :C 1))'] = {'(foo a)\\n  C', 'no named fn docstring mismatch'},
+    ['(doc (λ foo [] :D 1))'] = {'(foo)\\n  D', 'no anonymous lambda docstring mistmatch'},
```
thatismatt: Typo "mistmatch" in a few places. Really minor, but I hope you agree it is worth catching before this get merged.

@jaawerth BTW this is awesome and I'm really looking forward to seeing it merged. Thanks!

technomancy:
If you want to push what you have for rendering destructure tables more nicely, I can fix the unmangling issues. Thanks!

jaawerth:
@technomancy all pushed! I actually finished the unmanging the other day and just wanted to update the tests further so that I didn't need to use the fake `doc_env` replacement for `doc` in the tests, and instead use `setfenv`/`debug.setupvalue` to mock `print` with something that just returns the string for that function.

Still, please do take a gander at the commits + final change just to make sure I didn't introduce any new bugs in the process of finishing this!

test.lua
jaawerth: Thanks! I actually got rid of that wording in the last update but I appreciate you calling it out.

jaawerth:
## as of this update
- Function names and arg syms are now unmangled. Any destructured table args appear as `<table>`
- Running `doc` on anonymous functions (those with a nil `:fnl/fn-name` metadata field) now displays as `<fn:anonymous>`
- Running `doc` on a `nil` docstring displays `(fnname arg1 arg2)\n  <docstring:nil>`
- The tests were reworded and reorganized/modified a bit

## changes not applied
- Making a `mock` function in the tests that uses `setfenv` or `debug.setlocal`, and use it to mock `print` for the `doc` function to just return the string (and removing the fake `doc_env` function in the tests
- serializing destructured table args

## changes worth consideration
- adding tests to ensure that, in future, metadata and docstrings remain disabled in the REPL unless enabled via `--metadata` or `opts.useMetadata` (and that `--no-metadata` continues to disable it even in the REPL)
- when metadata IS disabled, replacing `fennel.doc` with a "metadata currently disabled" function (or similar) rather than displaying `(<fn:anonymous>)\n  <docstring:nil>`

technomancy:
Thanks! So glad to see this merged.

PR #177, opened 2019-08-30
Use code blocks whenever possible and tag them as Fennel code

christoph-frick:


technomancy:
Hopefully some day we'll have Fennel syntax highlighting in GitHub, but for now having it in Pygments is a start!

christoph-frick:
This was more in preparation for the fennel-lang.org repo

PR #178, opened 2019-09-06
Decrease header level in API docs

christoph-frick:
This makes this document equal in header levels to reference.md and changelog.md


technomancy:
Nice catch; thanks.

Issue #179, opened 2019-09-11
finalize metadata + docstrings for release


jaawerth:
Extending the work introduced by #173 (see also #154)

- [x] Ensure that when metadata is not enabled, `(doc foo)` either displays a warning that metadata is disabled, or `doc` should not exist.
- [x] Ensure that calling `doc` on nil shows an error.
- [x] Add a separate function containing the retrieval/formatting docstring logic.
- [x] Add docstrings to special forms
- [x] Turn `doc` into a macro so it can access special forms
- [x] Create documentation for the final docstrings/metadata API
- [x] Modify the `fennel -h` text to mention that the REPL defaults to enabling metadata

jaawerth:
Unless something else comes up it looks like we're good on the docstrings front!

PR #181, opened 2019-09-15
Change doc to being a macro

technomancy:
This addresses a bunch of the things in #179.

Making it into a macro allows it to use the "given" name rather than the name a function/special was originally defined under, which fixes one of the TODOs. I've added docstrings for `doc` and `doto` but the rest of the built-in macros and specials still need documenting.

I had to introduce a new `moduleName` option to the compiler removes the restriction that Fennel must always be required as the `"fennel"` module; we needed this anyway because sometimes you'll load the module under another path, like in Polywell: https://git.sr.ht/~technomancy/polywell/tree/master/polywell/lib

@jaawerth thoughts?

### DIFF
--- a/fennel
+++ b/fennel
@@ -37,7 +37,8 @@ Run fennel, a lisp programming language for the Lua runtime.
   If ~/.fennelrc exists, loads it before launching a repl.]]
 
 local options = {
-    sourcemap = true
+    sourcemap = true,
+    moduleName = "fennel",
 }
 
 local function dosafe(filename, opts, args)
--- a/fennel
+++ b/fennel
@@ -155,8 +155,10 @@ if arg[1] == "--repl" or #arg == 0 then
         -- pass in options so fennerlrc can make changes to it
         dosafe(initFilename, options, options)
     end
-    print("Welcome to Fennel " .. fennel.version .. "!\n" ..
-              "Use (doc something) to view documentation.")
+    print("Welcome to Fennel " .. fennel.version .. "!")
+    if options.useMetadata ~= false then
+        print("Use (doc something) to view documentation.")
+    end
     fennel.repl(options)
 elseif arg[1] == "--compile" then
     for i = 2, #arg do
--- a/fennel
+++ b/fennel
@@ -38,7 +38,6 @@ Run fennel, a lisp programming language for the Lua runtime.
 
 local options = {
     sourcemap = true,
-    moduleName = "fennel",
 }
 
 local function dosafe(filename, opts, args)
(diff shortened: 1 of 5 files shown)


jaawerth:
Looks good! We discussed making `doc` a macro early on in the process, but opted to at least try to keep it as a function if we could. Per chat in IRC, though, we need it to be a macro to allow for docstrings in macros and specials anyway.

I really like the addition of `moduleName` - it was always a concern that being able to work with metadata required requiring fennel itself, and this will allow greater flexibility in situations where metadata is desired.

This does break the ability to get the `doc` from a function returned from another function without first binding it to a variable/symbol. I don't see folks needing to do this much, though.

jaawerth:
It occurs to me that if we want the metadata system to be extensible for other purposes: with `moduleName`it won't be as simple as working with `fennel.metadata` from user code.

We should probably add a macro or two that abstracts `moduleName`, which from what I can see here is only accessible in compiler scope, under the hood so people can get and set custom metadata, and perhaps provide a way to gain access to the whole metadata table.

technomancy:
@jaawerth I think you're right that we need a bit more to make the metadata system a bit more general-purpose, but maybe that can wait till after the next release? Do you think it will require changes that will introduce backwards-incompatibility?

technomancy:
Good call on adding a test for user-defined macros. It exposed a problem where the new macro compiler scope hadn't propagated the `useMetadata` flag. Pushed a fix so we should be good now.

jaawerth:
Nice!

Regarding adding macros for interacting with metadata, it probably won't require breaking changes unless we need to tweak how `moduleName` works as a result. And worst-case scenario we end up cutting `0.4.0` sooner than we'd hoped, which isn't the end of the world.

However, it looks like it might still need tweaking - if you require a module with documented functions from the REPL, `(doc foo)` isn't showing the argslists or docstring for it, probably because there are now two competing metadata stores.

That also brought to my attention that if you have an undocumented function (one without metadata), then you get this:

```
>> (doc foo)
(foo)
  <docstring:nil>
```

Are we comfortable with that, or do we want to denote something when the argslist is unknown?

technomancy:
> if you require a module with documented functions from the REPL, (doc foo) isn't showing the argslists or docstring for it, probably because there are now two competing metadata stores.

Are you sure this isn't just another case of options propagation getting dropped?

> do we want to denote something when the argslist is unknown?

Yeah, we definitely shouldn't imply it's a 0-arity function; good catch.

jaawerth:
Hrm, it also doesn't seem to work when you run `doc` on a function you've required from another module.

technomancy:
OK, I got a failing test for the final issue of docstrings coming from require; can't seem to get that useMetadata setting to propagate correctly. @jaawerth if you get a chance could you take a look?

PR #190, opened 2019-09-30
Marked things that changed in v0.3, in the documentation

decadent:
Rationale: since only the latest version of the docs is available at fennel-lang.org (currently), some people may visit it who are still using a previous version. It would be helpful toward them to note changes between versions right in the docs, so there's no need to bump into incompatibilities and then study the changelog. IMO this and/or docs versioning will need to be introduced sooner or later.

Of course, I don't know if the APIs change too often yet and it's too early for versioning, but that doesn't seem to be the case so far.

Not sure that I caught all changes in 0.3.

Didn't touch the tutorial since I suppose people use it with the latest Fennel version.

### DIFF
--- a/api.md
+++ b/api.md
@@ -10,7 +10,7 @@ usually accept these fields:
 * `correlate`: when this is truthy, Fennel attempts to emit Lua where the line
   numbers match up with the Fennel input code; useful for situation where code
   that isn't under your control will print the stack traces.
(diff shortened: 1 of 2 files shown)


technomancy:
> since only the latest version of the docs is available at fennel-lang.org

We actually have docs for every released tag now.

But in any case, this is still useful. Thanks!

Issue #187, opened 2019-09-21
Automatically load luarocks in order to activate repl completion?


technomancy:
On a Debian install with the `readline` luarock installed without sudo, completion doesn't work unless you launch Fennel using `lua -lluarocks.loader fennel`. (or unless you installed Fennel from Luarocks)

We should consider activating the luarocks loader when attempting to load the readline module if a normal require fails. I don't know much about luarocks so I'd welcome input from folks who might be aware of pitfalls here.

technomancy:
We can address this using documentation as part of #236.

PR #192, opened 2019-10-05
readline completion on table fields

### LINKED ISSUE #188: Completion should descend into tables
Right now completion only works on top level values in globals, locals, and specials.

If you try to complete `table.co<TAB>`, you get no results; it should complete to `table.concat`.

We can take the implementation from Polywell here that knows how to descend into table fields: https://git.sr.ht/~technomancy/polywell/tree/master/polywell/commands.fnl#L215

jaawerth:
Resolves #188 

## Details
If your current partial identifier contains `.`, it will recursively consume the leading text and see if it's a table, then start matching on children.

To make it easier to complete table paths, I got rid of the appended space upon successful completion (which was a default in [readline.lua](https://pjb.com.au/comp/lua/readline.html) that has to be overridden each time with `set_completion_append_character`).

I also simplified the way it's tokenizing the input (line 2328 in the diff) to obtain the inputFragment it will try to complete.

## Why no tests?

I was initially going to add some tests, but I realized we don't have any readline-related tests at all, so I think it makes more sense to add this along with any/all other tests we create as part of #187

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,6 +2,8 @@
 
 ## 0.4.0 / ???
 
+* allow REPL readline completion to descend into table fields when using dot
+accessor in identifier (#192)
 * ???
 
 ## 0.3.0 / 2019-09-22
--- a/fennel
+++ b/fennel
@@ -124,7 +124,13 @@ local function tryReadline(opts)
           completer = replCompleter
         end
         local function replCompleter(text, from, to)
-          if completer then return completer(text:sub(from, to)) else return {} end
+          if completer then
+            -- stop completed syms from adding extra space on match
+            readline.set_completion_append_character('')
+            return completer(text:sub(from, to))
+          else
+            return {}
+          end
         end
         readline.set_complete_function(replCompleter)
       end
--- a/fennel.lua
+++ b/fennel.lua
@@ -2325,21 +2325,35 @@ local function repl(options)
 
     local replCompleter = function(text)
         local matches = {}
-        local inputFragment = text:gsub("[%s)(]*(.+)", "%1")
+        local inputFragment = text:gsub(".*[%s)(]+", "")
 
-        -- adds any matching keys from the provided generator/iterator to matches
-        local function addMatchesFromGen(next, param, state)
-          for k in next, param, state do
+        -- adds partial key matches in tbl to the match list
+        local function addPartials(input, tbl, prefix)
+          for k in pairs(tbl) do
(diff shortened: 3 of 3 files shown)


jaawerth:
In working on this, I also noticed we could, if desired, probably move all that completion logic out of `fennel.lua` and back into `fennel` by having the completer use `fennel.eval` with either a macro or `eval-compiler` to access `_SCOPE`, `_SPECIALS`, etc, which might improve separation of concerns, but figured I'd tinker with that separately.

jaawerth:
Ok, did a little cleanup and readability changes. I think this is good to go, but wouldn't mind +1 before merging just as a sanity-check.

kenrestivo:
Cool! Will this work with fennel-mode.el?

jaawerth:
Good question! I'm not an emacs user so I wasn't even thinking along those lines, but it's definitely worth playing with it to see if we can't make it accessible to tooling other than readline.

In theory it should be usable as is with a little finagling, though.

technomancy:
@kenrestivo unfortunately fennel-mode works in a line-at-a-time way, so while we could use the exposed completion logic, it would take some plumbing. Emacs would need a way to send the current input to the subprocess, which would need some external channel to send the list of completions back.

technomancy:
@jaawerth nice work on this; thanks! I think it looks great.

technomancy:
> In working on this, I also noticed we could, if desired, probably move all that completion logic out of fennel.lua and back into fennel

I don't think this is a good idea; completion is a part of the repl; whether you use readline or some other interface. For instance, in polywell we still want access to the completion logic, but we replace readline with our own interface to supply the input and show the possible targets. So keeping that in `fennel.lua` is very helpful.

> I was initially going to add some tests, but I realized we don't have any readline-related tests at all

IMO completion is not a readline feature; it's a general feature that just happens to be exposed using readline in the default setup. So we shouldn't write tests for "what happens when you press tab", but having tests for "if I have input X with env Y, I should get results Z" would make plenty of sense.

jaawerth:
@technomancy very good points. Having some built-in completion logic exposed this way will make it easier to make it work with tooling, for sure, and you're right that it isn't inherent to readline itself.

In that case, when writing tests (and documenting the completion api) it might be worth considering simplifying the API by simply exposing the completion function on the fennel API itself rather than the callback registration dance I threw together to avoid it. We could even consider having it take an optional predicate for folks who might want to, say, fuzzy-match or grab every possible field to dump into some external completion function (like those in some editors).

PR #196, opened 2019-11-27
Fix completion for mangled syms & tables w/ complex keys

### LINKED ISSUE #195: make REPL completion show unmangled names for locals
I noticed the other day that if you set a snake-cased variable like `foo-bar`, completion shows the mangled `foo_bar` form, as I failed to account for local name mangling when I implemented completion.

This should on the face of it be simple enough to fix via `scope.manglings` in the REPL.

jaawerth:
- Fixes #195 
- Also fixes an additional bug I realized I introduced with deep table completion, where non-string keys on a table throw an error due to `k:sub(...)`.

I created a wrapper for the REPL so we could write tests against both the REPL itself and the completer. Right now it's just in `tests.lua`, but I think we should start to consider pulling some of those helpers out into a `test-utils.lua` module. Figured I'd submit this and we could look at doing that in another PR, though.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -2328,33 +2328,36 @@ local function repl(options)
         local inputFragment = text:gsub(".*[%s)(]+", "")
 
         -- adds partial key matches in tbl to the match list
-        local function addPartials(input, tbl, prefix, unmangle)
+        local function addPartials(input, tbl, prefix)
             for k in pairs(tbl) do
-                local rawKey = unmangle and unmangle[k] or k
+                if tbl == env or tbl == env.___replLocals___ then
+                    k = scope.unmanglings[k] or k
+                end
                 if #matches >= 40 then break -- cap completions at 40
-                elseif input == rawKey:sub(0, #input) then
-                    table.insert(matches, prefix .. rawKey)
+                elseif input == k:sub(0, #input) then
+                    table.insert(matches, prefix .. k)
                 end
             end
         end
         -- adds matches to the match list, descending into table fields
-        local function addMatches(input, tbl, prefix, unmangle)
+        local function addMatches(input, tbl, prefix)
             prefix = prefix and prefix .. "." or ""
             if not string.find(input, "%.") then -- no (more) dots, so add matches
-                return addPartials(input, tbl, prefix, unmangle)
+                return addPartials(input, tbl, prefix)
             end
             -- check for table access field.child, and if field is a table, recur
             local head, tail = string.match(input, "^([^.]+)%.(.*)")
-            local rawHead = unmangle and unmangle[head] or head -- check mangling
+            local rawHead = tbl == env or tbl == env.___replLocals___
+                and scope.manglings[head] or head -- check mangling
             if type(tbl[rawHead]) == "table" then
-                return addMatches(tail, tbl[rawHead], prefix .. rawHead)
+                return addMatches(tail, tbl[rawHead], prefix .. head)
             end
         end
 
         addMatches(inputFragment, scope.specials or {})
         addMatches(inputFragment, SPECIALS or {})
-        addMatches(inputFragment, env.___replLocals___ or {}, nil, scope.unmanglings)
-        addMatches(inputFragment, env, nil, scope.unmanglings)
+        addMatches(inputFragment, env.___replLocals___ or {}, nil)
+        addMatches(inputFragment, env, nil)
         addMatches(inputFragment, env._ENV or env._G or {})
         return matches
     end
--- a/fennel.lua
+++ b/fennel.lua
@@ -2356,8 +2356,8 @@ local function repl(options)
 
         addMatches(inputFragment, scope.specials or {})
         addMatches(inputFragment, SPECIALS or {})
-        addMatches(inputFragment, env.___replLocals___ or {}, nil)
-        addMatches(inputFragment, env, nil)
+        addMatches(inputFragment, env.___replLocals___ or {})
+        addMatches(inputFragment, env)
         addMatches(inputFragment, env._ENV or env._G or {})
         return matches
     end
--- a/fennel.lua
+++ b/fennel.lua
@@ -2334,7 +2334,7 @@ local function repl(options)
                     k = scope.unmanglings[k] or k
                 end
                 if #matches >= 40 then break -- cap completions at 40
-                elseif input == k:sub(0, #input) then
+                elseif type(k) == 'string' and input == k:sub(0, #input) then
                     table.insert(matches, prefix .. k)
                 end
             end
--- a/fennel.lua
+++ b/fennel.lua
@@ -2328,31 +2328,34 @@ local function repl(options)
         local inputFragment = text:gsub(".*[%s)(]+", "")
 
         -- adds partial key matches in tbl to the match list
-        local function addPartials(input, tbl, prefix)
-          for k in pairs(tbl) do
-            if #matches >= 40 then break -- cap completions at 40 to avoid overwhelming
-            elseif input == k:sub(0, #input) then
-              table.insert(matches, prefix .. k)
+        local function addPartials(input, tbl, prefix, unmangle)
+            for k in pairs(tbl) do
+                local rawKey = unmangle and unmangle[k] or k
+                if #matches >= 40 then break -- cap completions at 40
+                elseif input == rawKey:sub(0, #input) then
+                    table.insert(matches, prefix .. rawKey)
+                end
             end
-          end
         end
         -- adds matches to the match list, descending into table fields
-        local function addMatches(input, tbl, prefix)
-          prefix = prefix and prefix .. "." or ""
-          if not string.find(input, "%.") then -- no (more) dots, so add matches
-            return addPartials(input, tbl, prefix)
-          end
-          local head, tail = string.match(input, "^([^.]+)%.(.*)")
-          -- check for table access field.child, and if field is a table, recur
-          if type(tbl[head]) == "table" then
-            return addMatches(tail, tbl[head], prefix .. head)
-          end
+        local function addMatches(input, tbl, prefix, unmangle)
+            prefix = prefix and prefix .. "." or ""
+            if not string.find(input, "%.") then -- no (more) dots, so add matches
+                return addPartials(input, tbl, prefix, unmangle)
+            end
+            -- check for table access field.child, and if field is a table, recur
+            local head, tail = string.match(input, "^([^.]+)%.(.*)")
+            local rawHead = unmangle and unmangle[head] or head -- check mangling
+            if type(tbl[rawHead]) == "table" then
+                return addMatches(tail, tbl[rawHead], prefix .. rawHead)
+            end
         end
 
-        addMatches(inputFragment, env._ENV or env._G or {})
-        addMatches(inputFragment, env.___replLocals___ or {})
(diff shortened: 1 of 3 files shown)


jaawerth:
Er, forgot to run `make testall` before making the PR. Looks like my use of a coroutine in [wrapRepl](https://github.com/bakpakin/Fennel/blob/bcb3b6b1686f289c82f2f6efbf02ddc0c76ee544/test.lua#L786-L810) (so we can make tests against it without going through `io` by pausing the infinite loop) works fine in lua 5.3, 5.2, and luajit, but 5.1 is *very* unhappy about it

```
Parse error: attempt to yield across metamethod/C-call boundary
Parse error: attempt to yield across metamethod/C-call boundary
Parse error: attempt to yield across metamethod/C-call boundary
Parse error: attempt to yield across metamethod/C-call boundary
Parse error: attempt to yield across metamethod/C-call boundary
Parse error: attempt to yield across metamethod/C-call boundary
Parse error: attempt to yield across metamethod/C-call boundary
Parse error: attempt to yield across metamethod/C-call boundary
Parse error: attempt to yield across metamethod/C-call boundary
Parse error: attempt to yield across metamethod/C-call boundary
Parse error: attempt to yield across metamethod/C-call boundary
Parse error: attempt to yield across metamethod/C-call boundary
Parse error: attempt to yield across metamethod/C-call boundary
Parse error: attempt to yield across metamethod/C-call boundary
Parse error: attempt to yield across metamethod/C-call boundary
Parse error: attempt to yield across metamethod/C-call boundary
Parse error: attempt to yield across metamethod/C-call boundary
Parse error: attempt to yield across metamethod/C-call boundary
...
```

I'll have to dig in and see what kind of workarounds there are, either by eliminating whatever metamethod call or lua built-in is causing it (its not like I'm using any C bindings) or by running that test in a patched lua 5.1, since it's only the test wrapper causing this.

We may also want to consider making the repl in `fennel.lua` event-based instead of using an infinite loop, in which case the coroutine would no longer be necessary to wrap it - then the `fennel` cli could be the part making use of `io.read`. That would also let people use the built-in repl with whatever kind of IO/socket implenetation they like. That would be another PR, though, and a breaking API change.

jaawerth:
I'm not going to get it done today, but I think the simplest workaround will be to eschew the coroutine entirely, and instead just use the normal REPL operation (which goes through `io.read`/`io.write` by launching a child process from the tests with `io.popen` and interacting with it that way.

jaawerth:
@technomancy  reminded me that wouldn't work since `io.popen` is only read OR write, not both. I ended up skipping the REPL tests for non-LuaJIT Lua 5.1 as a workaround for now, as the error is caused by the test util, not the code itself.

Could use a sanity-check review when someone gets a chance, or I can wait a bit and do it when it's not as fresh. Apologies for the large diff - I corrected the indentation while I was at it.

test.lua
```diff
@@ -781,6 +781,68 @@ do
...
+            pp = function(x) return x end,
+        })
+    end) replSend()
+    return replSend, replComplete
+end
+
+-- Skip REPL tests in non-JIT Lua 5.1 only to avoid engine coroutine bug
+if _VERSION ~= 'Lua 5.1' or type(jit) == 'table' then
+    local send, comp = wrapRepl()
+    local v = {
```
technomancy: I don't understand the purpose of the `v` table here. Wouldn't it be simpler just to make `wrapRepl` return functions which are usable as-is?

test.lua
```diff
@@ -781,6 +781,68 @@ do
...
+        {v.comp'foo',  view{'foo', 'foo-bar*'},
+            'local completion works & accounts for mangling'},
+        {v.comp'moe-larry.', '[ "moe-larry.*curly*" ]',
+            'completion traverses tables without mangling keys when input is "tbl-var."'},
+        {v.send'(values !x-y !x_y)', view{{1, 2}},
+            'mangled locals do not collide'},
+        {v.comp'!x', view{'!x-y', '!x_y'},
+            'completions on mangled locals do not collide'},
+    }
+    do
```
technomancy: What's the purpose of the extra `do/end` here?

technomancy:
> We may also want to consider making the repl in fennel.lua event-based instead of using an infinite loop

I feel pretty strongly that we shouldn't do this--the whole point of coroutines is that you should be able to get the best of both worlds: code written as if it was running linearly but actually running in a way that allows it to relinquish flow control.

fennel.lua
```diff
@@ -2329,30 +2329,36 @@ local function repl(options)
...
+                return addMatches(tail, tbl[rawHead], prefix .. head)
+            end
         end
 
-        addMatches(inputFragment, env._ENV or env._G or {})
-        addMatches(inputFragment, env.___replLocals___ or {})
-        addMatches(inputFragment, SPECIALS or {})
         addMatches(inputFragment, scope.specials or {})
+        addMatches(inputFragment, SPECIALS or {})
+        addMatches(inputFragment, env.___replLocals___ or {}, nil)
```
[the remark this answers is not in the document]
jaawerth: Yep, that was left over from an earlier form where I was passing the manglings after the prefix, before I decided I should just access it directly. Nice catch!

test.lua
jaawerth: Just to provide scope isolation for the transient values (such as the wrapped REPL instance, which is ended by the `nil` write). Now that everything is wrapped in an `if... end` it's unnecessary, so I'll remove it.

jaawerth: Oh wait, no - in this case it's now part of the for loop that runs the test cases. Should I move the `do` up to the same line as the closing brace to make that clearer?

fennel.lua
```diff
@@ -2329,30 +2329,36 @@ local function repl(options)
...
         local function addPartials(input, tbl, prefix)
-          for k in pairs(tbl) do
-            if #matches >= 40 then break -- cap completions at 40 to avoid overwhelming
-            elseif input == k:sub(0, #input) then
-              table.insert(matches, prefix .. k)
+            for k in pairs(tbl) do
+                if tbl == env or tbl == env.___replLocals___ then
+                    k = scope.unmanglings[k] or k
+                end
+                if #matches >= 40 then break -- cap completions at 40
```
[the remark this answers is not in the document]
jaawerth: Yes, that occurred to me back when I hardcoded the limit too. I can add that in this PR if you'd like to save on updates but the hardcoded behavior is what was previously in completion so it would also make sense as a separate PR. I'm good either way on that front, I don't mind adding it to this one.

Another alternative would be to have the internal completer return a stateful iterator function that a consumer could call as many times as they'd like to form a table.

test.lua
```diff
@@ -781,6 +781,68 @@ do
...
+                table.insert(output, xs)
+            end,
+            registerCompleter = function(completer)
+                replComplete = function(input)
+                    local res = completer(input)
+                    table.sort(res)
+                    return res
+                end
+            end,
+            pp = function(x) return x end,
```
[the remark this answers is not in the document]
jaawerth: Initially I wanted `pp` to just return the values since we're testing it from within lua anyway, and then use the already-implemented `deep_equal` function(s) to check the values. The use of `view` later was a workaround due to the current `deep_equal` having an infinite recursion bug when certain table shapes are passed to it (as I recall, `{{'foo'} {'foo'}}` caused it).

Longer-term I thought it would make sense to fix `deep_equal` so I just wrote that up in the meantime, but I can just as easily leave `pp` as-is, you're right.

test.lua
jaawerth: Initially I wanted `wrapRepl` generalized - per my comment above, the `v` table is mostly just a workaround for `deep_equal` being broken, and later when it's fixed `wrapRepl` could be left the same and the `v` table just removed. I'm comfortable removing it now and just updating `wrapRepl` though.

jaawerth:
> > We may also want to consider making the repl in fennel.lua event-based instead of using an infinite loop
> 
> I feel pretty strongly that we shouldn't do this--the whole point of coroutines is that you should be able to get the best of both worlds: code written as if it was running linearly but actually running in a way that allows it to relinquish flow control.

True. My main thought was that it if it was rewritten using tail calls, you could wrap the REPL without needing coroutines, which would avoid Lua 5.1 bugs like the one I initially hit here. But since it's unlikely for folks to use 5.1 outside of luajit, that may be a bunch of unnecessary re-implementation without much real benefit - I don't feel strongly about it, just figured I'd mention it.

test.lua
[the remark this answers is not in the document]
jaawerth: Looking at it in context (and considering the later discussion about just having the wrapped REPL return usable values), I think it still makes sense to replace `pp` just to avoid double-serialization, since the REPL can return multiple values.

fennel.lua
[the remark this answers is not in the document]
technomancy: Yeah, let's push this off to a separate PR. An iterator would be more flexible, but I don't think the extra complexity is called for.

test.lua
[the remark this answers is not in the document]
technomancy: Hm; I can't reproduce the problem with `deep_equal`.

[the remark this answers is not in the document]
jaawerth: Hrm, either I somehow fabricated this or it isn't happening with the tests as-is. Either way, it's cleaner with `deep_equal` so I'm going to push a version with the serialization stuff removed. Thanks for the sanity-check!

test.lua
jaawerth: Removing the `v` table!

test.lua
technomancy: I guess it's just that we don't typically worry much about isolating these scopes ; if we re-declare a local with the same name later on, it'll get caught by luacheck, so it just stands out a bit here. But it's fine.

test.lua
```diff
@@ -781,6 +781,62 @@ do
...
+            readChunk = function()
+                local chunk = coroutine.yield(output)
+                output = {}
+                return chunk and chunk .. '\n' or nil
+            end,
+            onValues = function(xs)
+                table.insert(output, xs)
+            end,
+            registerCompleter = function(completer)
+                replComplete = function(input)
```
technomancy: Shouldn't this just be `replComplete = completer`? Or are we specifically trying to get rid of multiple values?

test.lua
```diff
@@ -781,6 +781,62 @@ do
...
+end
+
+-- Skip REPL tests in non-JIT Lua 5.1 only to avoid engine coroutine bug
+if _VERSION ~= 'Lua 5.1' or type(jit) == 'table' then
+    local send, comp = wrapRepl()
+
+    send('(local [foo foo-bar* moe-larry] [1 2 {:*curly* "Why soitenly!"}])')
+    send('(local [!x-y !x_y] [1 2])')
+    for _, results in ipairs{
+        {comp'foo',  {'foo', 'foo-bar*'},
```
technomancy: Minor thing but the way the parens are omitted here bugs me; I think it'd be more readable with them added back.

jaawerth: :+1:  I admittedly went back and forth on that one, shoulda listened to my inner voice.

test.lua
jaawerth: ...yup. Poor form on my part, I'd added that when serializing when so I could sort first, since I'd noticed the order was wrong when serializing, but forgot to remove the wrapper when I changed it back. D'oh

technomancy:
Looks great; thanks!

PR #198, opened 2019-11-30
Fix windows batch path

Luxed:
**Issue:**
When using Windows and adding the git repo to the Path, an error occurs because windows cannot find `fennel`.

**Fix:**
To fix this, I used `%~dp0` before `fennel`. This means that the drive letter and script path is added before `fennel`, forcing it to be the complete path.


bakpakin:
Thanks, I guess we had not many windows users with fennel.

Issue #197, opened 2019-11-29
global special can collide with locals when global sym is valid lua with _


jaawerth:
It appears that globals created with the `global` special only mangle when the symbol doesn't contain a valid lua identifier. Based on prior issues this appears to be deliberate, but it can lead to reassignment to locals in some cases.

For example, because `_` is valid in lua identifiers and skips mangling and other checks, leading to a collision that will overwrite a local:

```clojure
(local foo-bar 1)
(global foo_bar 9999)
(print foo-bar)
```
compiles to
```lua
local foo_bar = 1
foo_bar = 9999
return print(foo_bar)
```

Whereas the following code correctly causes compiler error `runtime error: expected global, found var`:

```clojure
(local foo-bar 1)
(global foo-bar 9999)
(print foo-bar)
```

A simple solution would be to enable mangling for all globals, but it appears the conditional mangling was added on purpose to allow for Lua interop. Could we add some other special form for that purpose, like `global!` to skip mangling? Or simply check the global sym against both mangled and unmangled form of locals while compiling?

technomancy:
> Or simply check the global sym against both mangled and unmangled form of locals while compiling?

I think this is the way to go; it seems reasonable to make it an error if the mangled form already exists. It's almost always going to be unintentional, and raising it to the attention of the programmer will allow them to resolve the conflict instead of letting it go undetected.

> Could we add some other special form for that purpose, like global! to skip mangling?

We already have `set-forcibly!` for that, which should do the trick.

jaawerth:
> > Or simply check the global sym against both mangled and unmangled form of locals while compiling?
> 
> I think this is the way to go; it seems reasonable to make it an error if the mangled form already exists. It's almost always going to be unintentional, and raising it to the attention of the programmer will allow them to resolve the conflict instead of letting it go undetected.

Makes sense. Do we care about the inverse, though? Accidental shadowing? E.g.

```
(global a_b 1)
(local a-b 99999)
(print a_b)
```

PR #199, opened 2019-12-01
Check for conflict between globals and mangled locals.

technomancy:
I think the best thing to do in this case is to raise an error, because it's probably an accidental conflict. It's easy to rename the local if that global name was intentional.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -561,6 +561,18 @@ local function globalUnmangling(identifier)
     end
 end
 
+-- If there's a provided list of allowed globals, don't let references thru that
+-- aren't on the list. This list is set at the compiler entry points of compile
+-- and compileStream.
+local allowedGlobals
+
+local function globalAllowed(name)
+    if not allowedGlobals then return true end
+    for _, g in ipairs(allowedGlobals) do
+        if g == name then return true end
+    end
+end
+
 -- Creates a symbol from a string by mangling it.
 -- ensures that the generated symbol is unique
 -- if the input string is unique in the scope.
@@ -649,18 +661,6 @@ local function declareLocal(symbol, meta, scope, ast)
     local mangling = localMangling(name, scope, ast)
     scope.symmeta[name] = meta
     return mangling
-end
-
--- If there's a provided list of allowed globals, don't let references
--- thru that aren't on the list. This list is set at the compiler
--- entry points of compile and compileStream.
-local allowedGlobals
-
-local function globalAllowed(name)
-    if not allowedGlobals then return true end
-    for _, g in ipairs(allowedGlobals) do
-        if g == name then return true end
-    end
 end
 
 -- Convert symbol to Lua code. Will only work for local symbols
@@ -1114,6 +1114,16 @@ local function destructure(to, from, ast, scope, parent, opts)
                 assertCompile(not (meta and not meta.var),
                     'expected local var', up1)
             end
+            if forceglobal then
+                assertCompile(not scope.unmanglings[raw],
+                              "global " .. raw .. " conflicts with local", ast)
+                scope.manglings[raw] = globalMangling(raw)
+                scope.unmanglings[globalMangling(raw)] = raw
+                if allowedGlobals then
+                    table.insert(allowedGlobals, raw)
+                end
+            end
+
             return symbolToExpression(symbol, scope)[1]
         end
     end
@@ -1461,12 +1471,6 @@ docSpecial('.', {'tbl', 'key1', '...'},
 
 SPECIALS['global'] = function(ast, scope, parent)
     assertCompile(#ast == 3, "expected name and value", ast)
-    -- globals tracking doesn't currently work with multi-values/destructuring
-    if allowedGlobals and isSym(ast[2]) then
-        for _,global in ipairs(isList(ast[2]) and ast[2] or {ast[2]}) do
-            table.insert(allowedGlobals, deref(global))
-        end
-    end
     destructure(ast[2], ast[3], ast, scope, parent, {
         nomulti = true,
         forceglobal = true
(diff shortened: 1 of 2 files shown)


fennel.lua
```diff
@@ -1461,6 +1461,8 @@ docSpecial('.', {'tbl', 'key1', '...'},
 
 SPECIALS['global'] = function(ast, scope, parent)
     assertCompile(#ast == 3, "expected name and value", ast)
+    assertCompile(not scope.unmanglings[tostring(ast[2])],
+                  "global " .. tostring(ast[2]) .. " conflicts with local", ast)
```
jaawerth: This looks like a good approach, but it won't catch it if there's destructuring in play. For example,
```clojure
(local a-b 1)
(global [a_b] [2])
```
bypasses the check.

technomancy: Nice catch! Updated to handle destructuring as well as do extra mangling in the reverse case as per the comment in #197.

fennel.lua
```diff
@@ -2814,7 +2817,7 @@ that argument name begins with ?."
                   (do (assert (not (. pattern (+ k 2)))
                               "expected rest argument in final position")
                       (table.insert bindings (. pattern (+ k 1)))
-                      (table.insert bindings [`(select ,k ((or unpack table.unpack)
+                      (table.insert bindings [`(select ,k ((or _G.unpack table.unpack)
```
jaawerth: I don't think `_G.unpack` is necessary here, since we're explicitly exposing `unpack` in the compiler scope to avoid having to account for it in macros and such.

Also, I think we want to account for scenarios where an environment is set without a `_G`, which is why we've avoided using it directly in the rest of the compiler?

fennel.lua
```diff
@@ -582,7 +594,8 @@ local function localMangling(str, scope, ast)
     end)
 
     local raw = mangling
-    while scope.unmanglings[mangling] do
+    while scope.unmanglings[mangling] or (allowedGlobals and
+                                          globalAllowed(mangling)) do
```
jaawerth: Correct me if I'm wrong, but I think this will fail to account for the shadowing issue when `allowedGlobals` is false, because globals that exist as valid lua symbols aren't entered into the manglings table. The test you added on line 175 wouldn't account for this, since it's just using the default `allowedGlobals`

I suspect the only way to account for this with 100% reliability is to have mangling also add a prefix of some sort, similar to the way global manglings does. I don't know if that's something we want to avoid or not, but since you can't know what globals exist during runtime it's the only surefire approach. Not that it would be foolproof, but someone would at least have to work pretty hard to break it.

Another alternative, which would add fewer prefixes all over the place but still allow accidental shadowing when doing lua interop (or with globals created with `(tset _G foo bar)` would be to enter all globals created via `global` to the manglings/unmanglings table, whether they've been mangled or not.

test.lua
```diff
@@ -241,7 +243,7 @@ local cases = {
         ["(let [(a b c d e f g) (if (= (+ 1 1) 2) (values 1 2 3 4 5 6 7))] (+ a b c d e f g))"]=28,
         -- IIFE in if statement required v2
         ["(let [(a b c d e f g) (if (= (+ 1 1) 3) nil\
-                                       ((or unpack table.unpack) [1 2 3 4 5 6 7]))]\
+                                       ((or _G.unpack table.unpack) [1 2 3 4 5 6 7]))]\
```
jaawerth: What was the rationale for adding the `_G` here? Not a big deal, it just doesn't look necessary since this test is running with the default `allowedGlobals`.

fennel.lua
jaawerth: ...it occurs to me that this is intentional - and that this approach still works so long as we're comfortable with `{ allowedGlobals = false }` means losing any protection against mangling causing accidental shadowing, since fennel can't know ahead of time what the globals will be.

If so, I suggest we make sure that's made clear in the docs/cli help.

fennel.lua
jaawerth: Nice, I thought addressing destructuring was going to end up messier - *and* it looks like you ended up addressing another unfiled bug in the process with the destructuring + global tracking!

fennel.lua
technomancy: The funny thing is without this you can't use `match` when you have strict globals mode turned on in 5.3+, because `(or unpack table.unpack)` is included in the macroexpansion.

test.lua
technomancy: Same as above; this is now needed because we removed `allowGlobals = false` from the eval call that the tests run under.

fennel.lua
technomancy: Yes, basically this means that the checking we do to ensure that locals and globals don't collide is contingent on not disabling strict local checking.

> enter all globals created via global to the manglings/unmanglings table, whether they've been mangled or not.

Interesting; I hadn't considered this approach, but it does allow us to apply the conflict checks more widely. I'll give it a shot tomorrow; thanks for the suggestion. Maybe the trade-off I made isn't strictly necessary.

fennel.lua
jaawerth: Huh, bit of a catch-22 there. On the one hand it'll break with strict globals, on the other hand it'll break in runtime if run in an environment that doesn't include `_G` itself. I suppose the latter is pretty rare, but I can't help but wish we could have it both ways!

fennel.lua
technomancy: Cool; I think that did the trick! Thanks for the suggestion to add the globals to the manglings; that's a more thorough fix for the problem.

test.lua
```diff
@@ -274,10 +276,6 @@ local cases = {
         -- macros with mangled names
         ["(require-macros \"test-macros\")\
           (->1 9 (+ 2) (* 11))"]=121,
-        -- macros loaded in function scope shouldn't leak to other functions
-        ["((fn [] (require-macros \"test-macros\") (global x1 (->1 99 (+ 31)))))\
-          (pcall (fn [] (global x1 (->1 23 (+ 1)))))\
-          x1"]=130,
```
jaawerth: @technomancy Sorry for not asking this before you merged, but just to be sure:
How come this test was dropped? I'm guessing I'm just not seeing where it's redundant but I wanted to doublecheck just in case it was an accident.

Your updates look great btw!

technomancy: Oh, it wasn't dropped; I just had to move it down to the compiler errors section because I turned strict global checks on. It's just that it won't compile any more.

jaawerth: D'oh! I swear I looked for it being added elsewhere. Whoops!

PR #200, opened 2019-12-01
Fix bug where if with implicit nil else wasn't binding target local.

### LINKED ISSUE #191: implicit if value does not bind to nil
Overriding variables that contain `?` or `!` is broken
=========================================

This is the code to explain it better:
```
(fn mprint [a b]
  (print (.. (tostring a) " " (tostring b))))

(fn main1 []
  (let [a :hello
        b? :world]                                                                     
    (let [(a b?) (if (and true false)
                     (values :a :b)
                     (or nil false)
                     (values :c :d))]
      (mprint a b?))))

(fn main2 []
  (let [a :hello
        b? :world]
    (let [[a b?] (if (and true false)
                     [:a :b]
                     (or nil false)
                     [:c :d]
                     [a b?])]
      (mprint a b?))))

(fn main3 []
  (let [a :hello
        b :world]
    (let [(a b) (if (and true false)
                    (values :a :b)
                    (or nil false)
                    (values :c :d))]
      (mprint a b))))

(main1)  ; hello nil
(main2)  ; hello world
(main3)  ; hello world
```
The first function prints `nil` as the second parameter because it gets set to `nil` at the top of the do block.  This is the output in `lua`:
```
local function main1()
  local a = "hello"
  local b_3f = "world"
  do
    local a, b_3f = a, nil
    if (true and false) then
      a, b_3f = "a", "b"
    elseif (nil or false) then
      a, b_3f = "c", "d"
    end 
    return mprint(a, b_3f)
  end
end 
```

technomancy:
An `if` that's having its value bound to a local must set that local no
matter what. Previously the local would never be set if the conditions
all were falsey and the else clause was an implicit nil; now we check
that case and set it as well.

Fixes #191.

@bakpakin I could use some review here; I'm not sure that I'm using `innerTarget` correctly. There could be some subtlety about it that I'm overlooking.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -1636,6 +1636,9 @@ SPECIALS['if'] = function(ast, scope, parent, opts)
             if hasElse then
                 emit(lastBuffer, 'else', ast)
                 emit(lastBuffer, elseBranch.chunk, ast)
+            elseif(innerTarget) then
+                emit(lastBuffer, 'else', ast)
+                emit(lastBuffer, ("%s = nil"):format(innerTarget), ast)
             end
             emit(lastBuffer, 'end', ast)
         elseif not branches[i + 1].nested then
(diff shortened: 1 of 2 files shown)


technomancy:
Great; thanks!

PR #202, opened 2019-12-07
Add optional unused local checking with --check-unused-locals flag.

technomancy:
This ended up being quite simple since the `hashfn` implementation already added `.used` to symbol metadata.


technomancy:
Thanks!

Issue #193, opened 2019-10-07
improving .fennelrc + cli


jaawerth:
Right now, `.fennelrc` doesn't allow you to do some things you'd want to do in an rc file, such as modifying `fennel.path`. Here I'm proposing a few improvements.

## TODOs

When invoking fennelrc,

- [x] add a CLI flag (perhaps `-u` or `-C` ?) to load a custom init file in lieu of `.fennelrc`, or to not load one at all
- [x] set both the `options` and `fennel` itself on the `env` so `fennel.path` and other things can be manipulated from the RC file
- [ ] when launching a repl from the CLI, pass some kind of mode/context flag to `.fennelrc` so it can differentiate repl usage vs running or compiling
- [ ] have the CLI use `.fennelrc` in every context in which it's invoked (or at least for execution - compiling is more debatable), rather than only in the repl

## Exploration/possible todos 
- Allow the use of `XDG_CONFIG_HOME` (see [XDG base directory specification](https://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html)) to allow other places to check for the init file, configurable via non-fennel-specific env vars, before falling back to `.fennelrc`

jaawerth:
In addition to the above, I've also been mulling this over, but I'm interested in whether people think this is overkill.

## possible changes to where `.fennelrc` is sourced

Right now it just looks in `$HOME/.fennelrc` by default. We may want to consider having it follow XDG behaviors, looking in `${XDG_CONFIG_HOME:-$HOME/.config/fennel/fennelrc}` before falling back to `$HOME/.fennelrc`.

It's also worth considering having it look for the file in the project root and scanning upwards, so project-specific fennelrc files can be used, but that might be going too far into territory better covered by a separate, more complex CLI tool than we want for the base lang. E.g. `fnl` tool ala clojure's [clj](https://clojure.org/guides/deps_and_cli).

technomancy:
I like all of these suggestions except the last one; I feel pretty strongly that project-specific stuff should not be included in Fennel itself without a lot more consideration.

For the first one, maybe a better flag would be to skip the loading of `.fennelrc` altogether instead of providing a different path?

jaawerth:
[...]

PR #207, opened 2019-12-16
Improve XDG-compliancy

TmpFinger1:
- XDG_CONFIG_HOME is not meant to be always set,
  only when overriding the default - $HOME/.config
  So for most users the path would expand to a non-existing /.fennelrc
- a program should have its own subdir under HOME/.config
- fix a regression - since $HOME/.fennelrc went into initFilename2,
  it wasn't actually loaded anymore. The code would try to load
  from the XDG location (/.fennelrc) and crash.

(Addresses part of #193)

### DIFF
--- a/fennel
+++ b/fennel
@@ -162,9 +162,14 @@ if arg[1] == "--repl" or #arg == 0 then
     tryReadline(options)
 
     if options.fennelrc ~= false then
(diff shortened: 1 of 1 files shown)


technomancy:
Nice catch; thanks. That definitely makes sense.

PR #212, opened 2019-12-23
Don't double-emit `else` when checking innerTarget

### LINKED ISSUE #211: in 0.3.1, `match` emits bad lua when the result is assigned
## Repro

Consider the following fennel code:
```clojure
(local foo (match x
             1 :one
             2 :two
             _ "NOT ONE OR TWO"))
```
### Expected output
This is the output in 0.3.0, which is valid, and expected in 0.3.1:
```lua
local x_text = nil
do
  local _0_0 = arg
  if (_0_0 == 0) then
    x_text = "ZERO"
  elseif (_0_0 == 1) then
    x_text = "ONE"
  else
    local _ = _0_0
    x_text = "NOT BINARY"
  end
end
return x_text
```
### Actual output
In 0.3.1, we get this invalid lua:
```lua
local x_text = nil
do
  local _0_0 = arg
  if (_0_0 == 0) then
    x_text = "ZERO"
  elseif (_0_0 == 1) then
    x_text = "ONE"
  else
    local _ = _0_0
    x_text = "NOT BINARY"
  else
  x_text = nil
  end
end
return x_text
```

jaawerth:
Fixes #211

This is a bit of a quick fix - it looks like the logic for `if` should, longterm, be consolidated into using `hasElse` for everything, which will make this more maintainable. For now, though, this should be enough to cut a `0.3.1-1` or `0.3.2`.

If we want to fix the bug before 0.4.0, we'll likely need to cut a separate bugfix release by also merging this onto `0.3.1`, before the addition of kvmap.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -1664,7 +1664,8 @@ SPECIALS['if'] = function(ast, scope, parent, opts)
             if hasElse then
                 emit(lastBuffer, 'else', ast)
                 emit(lastBuffer, elseBranch.chunk, ast)
-            elseif(innerTarget) then
+            -- TODO: Consolidate use of condLine ~= "else" with hasElse
+            elseif(innerTarget and condLine ~= 'else') then
                 emit(lastBuffer, 'else', ast)
                 emit(lastBuffer, ("%s = nil"):format(innerTarget), ast)
             end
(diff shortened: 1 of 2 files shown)


jaawerth:
@technomancy since this may require a new release after applying this commit to the tip of `0.3.1`, I'm going to assign this to you rather than just merging, even though it's a pretty straightforward change.

technomancy:
Thanks; I'll cut a separate branch with this backported to it.

PR #215, opened 2019-12-28
re-implement include to use package.preload

### LINKED ISSUE #214: include emits bad output when nested
## Test case

Consider the following code, compiled with `--require-as-include` (which behaves the same if you change `require` to `include` and compile without that flag). It runs fine when invoked normally, but errors with `--require-as-include`:
**foo.fnl**
```clojure
(local foo [:FOO 1])
(local bar (require :bar))
(print (.. "foo: " (table.concat foo " - ")))
(print (.. "bar: " (table.concat bar " - ")))
```
**bar.fnl**
```clojure
(local bar [:BAR 2])
(each [_ v (ipairs (require :baz))]
  (table.insert bar v))
bar
```
**baz.fnl**
```clojure
[:BAZ 3]
```
The output with `--require-as-include` is:
```lua
local foo = {"FOO", 1}
local bar = bar
local _0_
do
local _0_
do
  _0_ = {"BAZ", 3}
end
  local bar = {"BAR", 2}
  for _, v in ipairs(_0_) do
    table.insert(bar, v)
  end
  _0_ = bar
end
bar = _0_
print(("foo: " .. table.concat(foo, " - ")))
return print(("bar: " .. table.concat(bar, " - ")))
```

## Solution: is include meant to be nested/repeated?

Now, even if we fix that output, if include naively inlines everything, that's going to lead to unexpected behavior, as you'll end up with two "modules" with totally separate state. So, I see two possible solutions here:

### 1. Bundle, or utilize `package.preload` (my preference)

I already have an implementation ready to PR for this. It turns `include` into a macro that inlines the code, attaches it to `package.preload`, and then invokes `require` normally. It works because `package.preload` is basically built for this purpose - one of the first searches in `package.loaders`/`package.searchers` checks for `package.preload[modulename]` and invokes it to load the module when present.

The only real downside to this approach is it won't work in Lua environments that don't have `package.preload`, but I don't know of any off the top of my head. I still think it would be better than the current behavior, which is broken, or a pure inlining behavior, as modules would always be correctly resolved.

### 2. Make include error when invoked on the same module more than once
This would break `--require-as-include` a lot, as it means you can't actually use it the same way you would require. It would really only be useful at a top-level entrypoint, pulling in a handful of 2nd-tier files. It would, however, be simpler, and could be used in any environment.

jaawerth:
Fixes #214 

This re-implements the `include` special as a macro that utilizes `package.preload` and `require`. It could use a heavier than normal review since it changes the underlying functionality of how it works so `include` and `--require-as-include` will work like require for a user.

@bakpakin, you may want to review this one since you implemented the original `include` and this makes quite a few changes to how it works.

## Notes
- Will use `load`/`loadstring` on a string literal instead of inlining the module as a closure. This could be changed back to a function wrapper, but this way it won't get upvalues from the parent scope.
- Got rid of `opts.fallback` and instead uses require with a simple `(values require)` to emit require and avoid infinite recursion when `options.requireAsInclude` is enabled.
- Adds a `get-option` function to the compiler/macro env, allowing macros to inspect options. I also considered a `get-options` instead that returns a shallow copy of the entire options table - could have gone either way here, and input is welcome.

## Caveats
While this will work in nested context, and will also ensure referential equality is correct. There are still a couple of edge cases though:
- When requiring modules that in turn require modules that have already been `include`ed by this macro, it will grab the one that's already been loaded. Similarly, this checks `package.loaded` and `package.preload` to avoid attempting to include something already loaded by Lua. However, it will still result in larger bundles due to the embedded (but unused) module text.
- The compiler output runs but it's a bit ugly due to the way it handles line-breaks.
- Not sure of the `(or loadstring load)` is necessary? I saw it falling back to load in `loadCode` in the compiler, so duplicated that here, but it looks like Lua 5.1-5.3 and LuaJit all have `loadstring` so this could potentially be removed.
- No tests. There aren't any tests yet for the `include` macro, and I didn't add any here yet. We may want to consider doing so?

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -468,7 +468,7 @@ local scopedInternals = setmetatable({}, {__mode = 'k'})
 -- 'eval-compiler' special form (may change). They use metatables to implement
 -- nesting via metatables.
 local function makeScope(parent)
-    return {
+    local scope = {
         unmanglings = setmetatable({}, {
             __index = parent and parent.unmanglings
         }),
@@ -490,6 +490,8 @@ local function makeScope(parent)
         depth = parent and ((parent.depth or 0) + 1) or 0,
         hashfn = parent and parent.hashfn
     }
+    scopedInternals[scope] = scopedInternals[scope] or {}
+    return scope
 end
 
 -- Assert a condition and raise a compile error with line numbers. The ast arg
@@ -1989,7 +1991,6 @@ local function macroToSpecial(mac)
     return special
 end
 
-local requireSpecial
 local function compile(ast, options)
     local opts = copy(options)
     local oldGlobals = allowedGlobals
@@ -2000,12 +2001,13 @@ local function compile(ast, options)
     if opts.indent == nil then opts.indent = '  ' end
     local chunk = {}
     local scope = opts.scope or makeScope(GLOBAL_SCOPE)
-    scopedInternals[scope] = scopedInternals[scope] or {}
     scopedInternals[scope].options = opts
     rootChunk = chunk
     rootScope = scope
     rootOptions = opts
-    if opts.requireAsInclude then scope.specials.require = requireSpecial end
+    if opts.requireAsInclude then
+        scope.specials.require = scope.specials.include
+    end
     local exprs = compile1(ast, scope, chunk, {tail = true})
     keepSideEffects(exprs, chunk, nil, ast)
     allowedGlobals = oldGlobals
@@ -2104,9 +2106,10 @@ local function compileStream(strm, options)
     allowedGlobals = opts.allowedGlobals
     if opts.indent == nil then opts.indent = '  ' end
     local scope = opts.scope or makeScope(GLOBAL_SCOPE)
-    scopedInternals[scope] = scopedInternals[scope] or {}
     scopedInternals[scope].options = opts
-    if opts.requireAsInclude then scope.specials.require = requireSpecial end
+    if opts.requireAsInclude then
+        scope.specials.require = scope.specials.include
+    end
     local vals = {}
     for ok, val in parser(strm, opts.filename) do
         if not ok then break end
@@ -2575,10 +2578,6 @@ docSpecial('require-macros', {'macro-module-name'},
            'Load given module and use its contents as macro definitions in current scope.'
                ..'\nMacro module should return a table of macro functions with string keys.')
 
-requireSpecial = function (ast, scope, parent, opts)
-    return SPECIALS['include'](ast, scope, parent, opts)
-end
-
 local function evalCompiler(ast, scope, parent)
     local luaSource = compile(ast, { scope = makeScope(COMPILER_SCOPE),
                                      useMetadata = rootOptions.useMetadata })
@@ -2826,7 +2825,7 @@ Macro module should return a table of macro functions with string keys."
                                 (tset package.preload ,module-name
                                       ((or loadstring load) ,lua-code)))]
      ; if requireAsInclude is enabled when unresolvable, fallback to require
-     (assert (or found (get-option (get-scope) :requireAsInclude))
+     (assert (or found (get-option scope :requireAsInclude))
              (.. "could not find module " module-name))
      (local out `(do ((values require) ,module-name)))
      (if lua-code (do (tset scope.includes module-name mod-path)
--- a/fennel.lua
+++ b/fennel.lua
@@ -2498,9 +2498,11 @@ module.doc = doc
 module.searcher = module.makeSearcher()
 module.make_searcher = module.makeSearcher -- oops backwards compatibility
 
+local compilerEnvRootOpts = rootOptions
 local function makeCompilerEnv(ast, scope, parent)
     return setmetatable({
         -- State of compiler if needed
+        _ROOT_OPTIONS = compilerEnvRootOpts,
         _SCOPE = scope,
         _CHUNK = parent,
         _AST = ast,
@@ -2511,6 +2513,7 @@ local function makeCompilerEnv(ast, scope, parent)
         fennel = module,
         -- Useful for macros and meta programming. All of Fennel can be accessed
         -- via fennel.myfun, for example (fennel.eval "(print 1)").
+        searchpath = searchModule, -- package.searchpath 5.1 fallback
         list = list,
         sym = sym,
         unpack = unpack,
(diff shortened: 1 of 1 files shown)


fennel.lua
```diff
@@ -2868,6 +2804,33 @@ that argument name begins with ?."
...
+         lua-code        (if code (compile code))
+         found           (or (. scope.includes module-name) lua-code)
+         includer        `(when (and (not (. package.preload ,module-name))
+                                     (= nil (. package.loaded ,module-name)))
+                                (tset package.preload ,module-name
+                                      ((or loadstring load) ,lua-code)))]
+     ; if requireAsInclude is enabled when unresolvable, fallback to require
+     (assert (or found (get-option scope :requireAsInclude))
+             (.. "could not find module " module-name))
+     (local out `(do ((values require) ,module-name)))
```
technomancy: Can you add a comment about why `(values require)` is necessary here? I think I understand it now, but I won't if I come back to it later. =)

technomancy:
> also considered a get-options instead that returns a shallow copy of the entire options table

I prefer `get-option`; I don't think we should expose more than we need to.

> I saw it falling back to load in loadCode in the compiler, so duplicated that here

I believe that fallback is actually due to `setfenv` being removed in newer Lua versions; if you're not changing the environment then you should be good with `loadstring`

I haven't used `include` so while I've looked thru the patch, I'd feel better about merging this with tests.

bakpakin:
I think a better solution is to fix current implementation. Using loadstring is not as clean as actually pre-compiling the code, and I'm fairly confident the current version at some point in the past worked fine for the all of the above use cases. I considered package preload because but in the end did not use it, because there really isn't that much change in logic to just inline the code and memoize the includes. All the same book keeping is required.

The code output is kind of awful, and I really believe that we can fix the current implementation with minimal changes. Any major changes here should definitely add tests so we don't fall prey to the same problems again, and I really should have added tests the first time, mea culpa.

jaawerth:
I can definitely add tests to this (whether we end up going with it or not).

@bakpakin can you point me to the version where it worked? I tried it with `0.3.0`, which seems to be the earliest I have where it exists, and couldn't find it. As for caching, I could be misreading your original implementation, but it looks like it's caching it in the compiler but not in the generated code? The issue with that is it will break referential equality in the code output. I wasn't really sure where to cache it in runtime other than either a hidden global or using an existing caching mechanism, which in this case would be preload or package.loaded.

I'm open to fixing the current implementation, of course (and it would be a lot less ugly with a generated `[[ ]]`-style string), but I suspect there are going to be other edge cases with inlining, and I don't think they're surmountable without parsing Lua.

For example:
1. in `wat.fnl` say you define a local variable that has the same name as one an `included` Lua module tries to set globally, or
2. Someone defines a local in `wat.fnl` that happens to shadow a global like `assert`. If it's inlined, even with `setfenv`, that will override unless you also loop through the upvalues and use `debug.setupvalue`.

These edge cases aren't good ideas to do in code to begin with, but if the design goal is to make `include` work like `require` from the perspective of the caller, how do we get around these without `loadstring`?

bakpakin:
I think the shadowing can be worked around by surrounding things with `do` ... `end`, or an anonymous function, so included modules don't "leak" their locals.  The only new symbols exposed by inlining would be a gensym. Also, inlined modules are prefixed to the beginning of the output file, so as not to be subject to any "contamination" by the parent module (this is how I intended the current, buggy implementation to work). As for globals, manual inling would change the order of execution (modules may be evaluated in a different order than they are normally evaluated in by require), but package.preload could fix this problem.

It seems that the original implementation probably did have some large issues here, not sure what went wrong. I remember my original test files working with multiple levels of include.

I'm open to merging this PR, if the (intended) semantics are kept and the code generation is improved.

jaawerth:
Hoisting the inlining to the top of the file makes sense - I kind of kicked myself after I edited my reply for the nth time because I belatedly realized it. We'd still need a runtime caching of **some** sort, which I think makes sense as package.preload, since otherwise we'd only get correct behavior with `--require-as-include` (sans any fallbacks, so it's all just one big file with the inlined stuff at the top), and only when there are no Lua modules that try to `require` something that's already been `include`d.

I think a reasonable way to combine the implementations might be:
1. No `loadstring`, but instead simply wrap the inlined code in a function so it only actually executes on demand
2. Set said function on `package.preload`
3. Hoist both 1 and 2 to the top of the generated file to avoid leakage from the parent module

Thoughts?

bakpakin:
I think that should pretty much cover everything. Keep in mind, if we eagerly evaluate everything, there is no need to keep a runtime cache, as `(include :module)` willl be replaced with some local in the top level scope, say `module_0_`. But that does fall prey to changing the order of evaluation, which is a subtle but perhaps important change.

I have a fix for the current logic locally (diffed from master), which simply patches how it was supposed to  work so I could verify to myself that I wasn't crazy and the original idea could work. (so no package.preload)

Here is the diff:
https://gist.github.com/bakpakin/713da783edf8bf2a894d957b5cd1c6bf

Given the original example, this generates code like the following:
```lua
local module_1_
local bar = bar
do
  module_1_ = {"BAZ", 3}
end
local module_0_
do
  local bar = {"BAR", 2}
  for _, v in ipairs(module_1_) do
    table.insert(bar, v)
  end
  module_0_ = bar
end
local foo = {"FOO", 1}
bar = module_0_
print(("foo: " .. table.concat(foo, " - ")))
return print(("bar: " .. table.concat(bar, " - ")))
```

By replacing `do` ... `end` and the generated `module_n_` locals with an anonymous function and package.preload, we would basically have your suggestion.

jaawerth:
Sounds good! I'll take a crack at adapting both this version and your diff to work that way and see where we land. I like the idea of most of it being a macro, maybe with a hook in `compile` that checks for `scope.includes` to emit at the top, as it would be nice if everything but the most atomic bits that need to generate Lua could be self-hosted AST transforms, but I could see either being preferable depending on how it plays out.

fennel.lua
jaawerth: Will do (if we end up keeping this form instead of using bakpakin's diff below)! That's actually what the comment at the top of this snippet was supposed to be account for, but I can see why it would be unclear due to both the position and wording..

bakpakin:
@jaawerth I'm going to push my current fix, along with a test case to master as I can't push to this branch (you will need to force push). I don't think this will mess with your changes and make it too hard to merge, and it will come with some test cases.

jaawerth:
@bakpakin @technomancy 

So I've made a version that builds on bakpakin's update that adds the use of `package.preload`. I'm a bit ambivalent about it - I really like the idea of simplifying this into a macro, but the part that works better as a special is splicing the source into the root chunk. I was thinking it would make sense to simply have a step in the compile process that iterates `rootScope.includes` and does the splicing then, but there are currently a lot of possible places to do this outside of a special, all sort of dirty, so it might make more sense to revisit turning this into a macro after we've finished updating the process to use the macro expander.

I also updated the tests to check for leakage of locals, since I noticed this is still happening in master when `include`ing a Lua source, largely due to only Fennel sources being added to the top.

Anyway, I haven't rebased and force-pushed this branch yet since what I ended up making is radically different. You can see the comparison in the [include-special-preload](https://github.com/bakpakin/Fennel/compare/include-special-preload?expand=1) branch.

Before I wipe out this branch with the other one, can either of you see a clean way to leave the macro implementation in place while still splicing the sources into `rootChunk`?

bakpakin:
@jaawerth, a traditional macro implementation of this is impossible, as the transform is non-local, while pure macro transformation must be local. I like the new implementation a lot, just  one thing.

Here:
```lua
local forms = list(sym('or'), targetForm, list(sym('fn'), sequence()))
```
I don't think the `or` form is needed, but perhaps you have some rational for that?

jaawerth:
@bakpakin ok, thanks, that's what I thought. I didn't think it was even doable without cheating a way of slipping a hacky operator function to only that one macro via that scopedInternals thig, which could also use the macro itself as a key in the weaktable or something like that. Didn't seem like a good enough rationale to include such a hack though. Definitely sounds like leaving it as-is (once we dicuss the tweaks) is the way to go for now.

Longterm, I have an idea for a hooks system we could use once we have things rebuilt around macro expansion but it's not worth leaving this open when we have a solution ready to go.

jaawerth:
Closing in favor of #217

PR #217, opened 2020-01-06
make `include` special use package.preload

### LINKED ISSUE #214: include emits bad output when nested

jaawerth:
Fixes #214
Alternative to #215 that:
- Leaves `include` as a special, for now (may revisit after we implement macro expansion as a compiler phase)
- Inlines the module code on the top of the entrypoint, wrapped in a function, ensuring no leakage of locals from the parent context into the wrapped code, and that execution order is maintained
- Sets the wrapped function on `package.preload` so that `require` can be used normally
- Emits `require` in place of `include`

## setting package.preload
Setting package.preload ensures execution order is maintained, and that duplicate and nested requires will have referential equality even if there's more than one entrypoint. What actually gets emitted is the equivalent of:
```lua
package.preload[modulename] = package.preload[modulename] or wrapped_function
```
This ensures that `include` maintains the same expectations as require, allowing one to override or mock the target module by setting it on `package.preload[modulename]`.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -2577,10 +2577,7 @@ SPECIALS['include'] = function(ast, scope, parent, opts)
     local mod = loadCode(code)()
 
     -- Check cache
-    local includeExpr = scope.includes[mod]
-    if includeExpr then
-        return includeExpr
-    end
+    if scope.includes[mod] then return scope.includes[mod] end
 
     -- Find path to source
     local path = searchModule(mod)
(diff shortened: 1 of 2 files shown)


bakpakin:
Looks good to me, merging. Thanks for all the work you put in for this fix, @jaawerth!

PR #219, opened 2020-01-07
fix outerTarget shadowing in let/do bindings

### LINKED ISSUE #218: broken shadowing with nested `let`/`do`
Thanks to @norcalli for pointing this out to me.

I think @technomancy and I got this pretty narrowed down for a fix already, but I wanted to get it filed for documentation purposes so we can get it into `0.3.2`.

## Repro

```clojure
(let [t (let [t {} k :a]
          (tset t k 123) t)
      k :b]
  (tset t k 321) t)
```
The above code should work fine with the shadowed `t`, but what we see in the output is:

```lua
local t = t
do
  local t = {}
  local k = "a"
  t[k] = 123
  t = t
end
local k = "b"
t[k] = 321
return t
```

## Fact-finding

Changing one of the bindings from `t` to `t2` unsurprisingly corrects this behavior. You get the same output if you change the `let`'s to `(do (local t {}) (local k :a) ..)` etc. However, using `do` with a destructured declaration works as expected:

```clojure
(do (local (t k) (values (do (local (t k) (values {} :a))
                             (tset t k 123)
                             t)
                         :b))
  (tset t k 321) t)
```
Just the outer destructure (the inner one isn't necessary for it to work) results in:
```lua
local t, k = t, k
local _0_
do
  local t, k = {}, "a"
  t[k] = 123
  _0_ = t
end
t, k = _0_, "b"
t[k] = 321
return t
```
This should narrow down which bit of the expression handling we need to stop from misbehaving.

jaawerth:
Partially addresses #218 
Added some code that checks for `outerTarget` in `subScope` and, if present, generates an intermediate target to assign to.

I initially did this in a separate conditional branch, but realized I could reuse the existing code that generates bindings when `opts.nval ~= 0`. It seems cleaner, but I left the commit in place just in case anyone wants to disagree.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -1279,17 +1279,18 @@ local function doImpl(ast, scope, parent, opts, start, chunk, subScope)
     subScope = subScope or makeScope(scope)
     chunk = chunk or {}
     local len = #ast
-    local outerTarget, finalTarget = opts.target
-    local outerTail = opts.tail
+    local outerTarget = opts.target
+    local finalTarget, outerTail = outerTarget, opts.tail
     local retexprs = {returned = true}
 
-    -- See if we need special handling to get the return values
-    -- of the do block
-    if not outerTarget and opts.nval ~= 0 and not outerTail then
-        if opts.nval then
+    -- if subScope shadows outerTarget, generate an intermediary
+    local badTarget = outerTarget and subScope.manglings[outerTarget]
+    -- See if we need special handling to get the return values of the do block
+    if badTarget or not outerTarget and opts.nval ~= 0 and not outerTail then
+        if badTarget or opts.nval then
             -- Generate a local target
             local syms = {}
-            for i = 1, opts.nval do
+            for i = 1, opts.nval or badTarget and 1 do
                 local s = gensym(scope)
                 syms[i] = s
                 retexprs[i] = expr(s, 'sym')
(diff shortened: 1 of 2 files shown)


jaawerth:
**Update:** I did some additional testing, and while this does address the shadowing issue for a single target, it doesn't account for multiple targets (multiple returns from `do`). I thought `nval` would be > 0 in that case and generate the intermediaries, but it looks like that scenario actually does an end run around the check, as `outerTarget` just becomes `x, y, z`. We'll likely need to split multiple bindings so we can check each of them separately.

jaawerth:
@bakpakin sorry to bug you, but I'm not super clear on this: should `nval` exist and be nonzero when trying to do something like the following?
```clojure
(local (x y) (do (local [x y] [1 2])
                 (values x y)))
(print x y)
```
The above is still broken under this PR - I've already implemented another test and a patch that would split `outerTarget` on `[^, ]` and use that as nval if nval isn't set, but that feels like a hacky workaround for a deeper problem.

jaawerth:
Upon further review,  the `subScope` check is pointless the way I've done it - it will always be true because of scope inheritance, and I can't use `rawget` because the shadowed declaration won't actually exist in `subScope` until *after* the body has been compiled, so this always uses an intermediate target when `outerTarget` exists.

We *could* go that way, and I could update this to not bother checking `subScope` at all. Assuming something isn't up with `nval`, the other option would be to try compiling the body first to get the scope, setting options accordingly, and compiling again. Or we could go another route and I can just close this PR.

bakpakin:
Hey, seems to be a bit confusion on what the nvals option does in the compiler, so let me try to explain.

In Lua, an expression can evaluate to 0 or more values via multiple returns. In many cases, Lua will drop extra values and convert a 0 value expression to nil. In other cases, Lua will use all of the values in an expression, such as in the last argument of a function call. Nval is an option passed to compile1 to say that the resulting expression should have at least n values. It let’s use generate better code, because if we know we are only going to use 1 or 2 values from an expression, we can create 1 or 2 locals to store intermediate results rather than turn the expression into a closure that is called immediately.

If we are compiling a `do` form, and the value of nval is 2, we should create 2 locals as our target, say `“_0_, _1_”` and assign to that.

technomancy:
Thanks for the detailed explanation; going to add this as a comment to the compiler.

PR #222, opened 2020-02-01
Switch to using luaunit instead of homegrown test thingy.

technomancy:
This makes our test suite *much* more organized.

Tests for fennelview and the compiler error messages are done in Fennel; the rest is left in Lua. The "core" tests retain their data-driven style, which means we still don't get line numbers for failures, but it's a start.

We don't necessarily have to stick with luaunit; lunatest or another alternative would be considered, but we talked on IRC about busted and there were some hesitations about how it organizes things. The main advantage of luaunit over lunatest is that it has better reporting when table structure doesn't match.

One thing I'm very much *not* keen on with luaunit is that it stops a test as soon as it encounters a failure; I haven't looked into whether that can be changed.


technomancy:
Thanks. Went ahead and ported a couple more suites to Fennel so now we're 4 and 4.

Issue #234, opened 2020-04-08
include with .lua files -- missing newline causes invalid code generation in some cases


XeroOl:
If `include` is used to include a file that doesn't have a trailing newline, in some cases, invalid code is generated.

For example: in `a.lua`:
```lua
print "this is my lua file"
-- comment
```
in `b.fnl`
```fnl
(include :b)
```
Compiling `b.fnl` generates:
```lua
package.preload["a"] = package.preload["a"] or function()
print "this is my lua file"
-- commentend
return require("a")
```
Although `a.lua` is a valid lua file, the generated code is invalid, because the `end` is commented out.
Fennel shouldn't trust that the included lua file will have a trailing newline.

XeroOl:
The fix for this is very simple, so I have created a pull request (#235) that adds the missing newline to the code generator, and changes a test case so that this issue is tested. Feel free to use or ignore that request.

jaawerth:
Nice catch!

PR #235, opened 2020-04-08
Fix issue #234

XeroOl:
This is a small change that should fix issue #234.


bakpakin:
LGTM, thanks for this fix!

PR #242, opened 2020-04-16
ensure that (values) emits valid lua 241 fix

### LINKED ISSUE #241: (values)
Currently, `(values)` doesn't emit correct code in some cases. I've found that `local`, `let`, and `var` can  combine with `(values)` to produce invalid code, but it's possible that there are other forms that cause similar issues. 
```fnl
(local t (values))
```
```lua
local t =
```
`lua:1:unexpected symbol near <eof>`

## Solutions

I've come up with a list of potential ways to solve this problem.

* Change `(values)` to be a compile error (such as "missing argument").
  * No easy way to avoid returning a value.
  * Breaking change that should appear in the changelog.
* Create a special case for no args, ie `(function()end)()`.
  * Potentially has performance implications
* Fiddle with the `nval` optimization so that the current code generation will be correct.
  * I suspect the issue is with the destructuring code.
* Leave the code as is and call `(values)` undefined behavior.
* Potentially something else?

XeroOl:
Should fix #241. I can't imagine this causing any other bugs. I've added a test specifically for empty values.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -964,9 +964,9 @@ local function handleCompileOpts(exprs, parent, opts, ast)
         emit(parent, ('return %s'):format(exprs1(exprs)), ast)
     end
     if opts.target then
-        local expr = exprs1(exprs)
-        if expr == '' then expr = 'nil' end
-        emit(parent, ('%s = %s'):format(opts.target, expr), ast)
+        local result = exprs1(exprs)
+        if result == '' then result = 'nil' end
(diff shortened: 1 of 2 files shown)


XeroOl:
```fnl
(let [a (values)
     b (values (values))
     (c d) (values)
     e (if (values) (values))
     f (while (values) (values))
     [g] [(values)]
     {: h} {:h (values)}]
     (not (or a b c d e f g h)))
```

technomancy:
Looks pretty thorough; thanks!

PR #246, opened 2020-04-24
Vararg limiting macros

jaawerth:
This adds two new macros, `limit-args` and `limit-values`.

## Why both?

Strictly speaking, we could probably get away with just `limit-values`, but the one that generates a new function was also in demand and covers a common use-case for wrapping functions.

Still, open to feedback on whether we really need both as part of the language.

### `(limit-values n ...)`

This operates on multiple values (passed directly, multiple returns from a function, `...`, etc) and returns the first n values only the first n. Useful in a number of scenarios to cap the values when passing around function invocations.

```clojure
(limit-values 0 :a :b :c :d :e) ; => nil
[(limit-values 2 (table.unpack [:a :b :c]))] ;-> ["a" "b"]

(fn sum-all [x y ...]
  (let [x (or x 0) y (or y 0)]
    (if (= (select :# ...) 0) (+ x y) (sum-all (+ x y) ...))))

(sum-all (limit-values 2 (values 10 10))) ; => 20
(->> [1 2 3 4 5] (table.unpack) (limit-values 3) (sum-all)) ; => 6
```

### `(limit-args n f)`

Like `limit-values`, but it wraps a function (or operator, or any expression) with a function that only passes along first n values. Useful for function composition, or for quickly doing `(fn [a b] (func a b))` for variadic functions that could receive bad input wen the varargs are left unchecked.

```clojure
(local sum-2 (limit-args 2 sum-all))
(sum-2 5 5 5 5) ; => 10
(-> [1 2 3 4 5] (table.unpack) ((limit-args 3 sum-all))) ; => 6
```

### DIFF
--- a/reference.md
+++ b/reference.md
@@ -134,9 +134,41 @@ Example:
 (partial (fn [x y] (print (+ x y))) 2)
 ```
 
-
 This example returns a function which will print a number that is 2
 greater than the argument it is passed.
+
+### `limit-values` limit varargs/multiple returns to the first n values
+
+Discard all values after the first n when dealing, with multi-values, `...`,
+and multiple returns. Useful for composing functions that return multiple values
+with variadic functions.
+
+Example:
+
+```fennel
+(limit-values 0 :a :b :c :d :e) ; => nil
+[(limit-values 2 (table.unpack [:a :b :c]))] ;-> ["a" "b"]
+
+(fn sum-all [x y ...]
+  (let [x (or x 0) y (or y 0)]
+    (if (= (select :# ...) 0) (+ x y) (sum-all (+ x y) ...))))
+
+(sum-all (limit-values 2 (values 10 10))) ; => 20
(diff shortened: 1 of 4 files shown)


reference.md
```diff
@@ -134,10 +134,42 @@ Example:
 (partial (fn [x y] (print (+ x y))) 2)
 ```
 
-
 This example returns a function which will print a number that is 2
 greater than the argument it is passed.
 
+### `limit-values` limit varargs/multiple returns to the first n values
+
+Discard all values after the first n when dealing, with multi-values, `...`,
```
Benaiah, line 142:
```suggestion
Discard all values after the first n when dealing with multi-values (`...`)
```

technomancy:
Looks great; nice work! Could we get a changelog mention too? Then I think we're set.

jaawerth:
Knew I was forgetting something. Done, thanks for looking it over!

PR #249, opened 2020-04-30
Friendly errors

technomancy:
This is just the start really, but I think it's a big improvement for usability when it comes to compiler errors:

```
Compile error in test/bad/set-local.fnl:2
  expected var x

(set x 3)
     ^
* Try declaring x using var instead of let/local.
* Try introducing a new local instead of changing the value of x.
```

I have chosen to skip introducing color for now because the errors are relatively short, and I don't think it benefits much from color to draw your eye to the `^^^` bits. But if the messages get longer, this might be more useful.

For a full catalog, see https://p.hagelb.org/fennel-friendly.html

## Implementation details

This is done as a new optional module called `fennelfriend` that is loaded inside a `pcall` and allowed to fail, much like `fennelview`. When it loads, it adds `options.assert-compile`/`options.parse-error`, and every time a compiler or parse error is thrown, we defer to those functions over the built-in dumber error printer.

Because this is optional, it means Fennel is still a one-file compiler.

Prior to this, we discarded file/line/offset data for table literals (with `[]` or `{}`) in the AST. However, this made pinpointing errors involving binding tables impossible. We can't just throw those fields directly into the tables like we do with lists, because they would get reflected in the compiled output. So instead I have put them on metatables on those nodes. It's a little more complex because now there are two places you have to look for source data, but I think it's justified. You can always ignore it and you're back to where we were before.

When we have a "proper" compiler plugin system, this system should be implemented in terms of that, but I don't think building the plugin system should block this useful feature.

## Bonus feature

This disables the stack traces upon normal compiler errors, since they are irrelevant to end users. If you're hacking the compiler and you want them back, `export FENNEL_DEBUG=trace`.

## Follow-up features

I have based this on a very cursory look at Elm. Most of the help that Elm gives depends on the rich type data their compiler has, and we will never be able to be as detailed as they are. But there may be other things they do that we should consider doing as well; further investigation is warranted.

This also assumes the relevant context consists only of a single line. There are probably cases where printing more of the source in question would be helpful.


technomancy:
I've extended this to cover parse errors now; I think it covers every known error that the compiler can throw.

technomancy:
If anyone wants to look thru the complete catalog of all errors here to give feedback that'd be great: https://p.hagelb.org/fennel-friendly.html

Otherwise I think this is good to go?

PR #254, opened 2020-05-02
Generate gensym for let body value before compiling binding vector.

### LINKED ISSUE #245: Wrong Lua generated for destructuring let inside tset
This:
```
(tset t :a (let [{: a} {:a :b}] a))
```
Compiles to:
```
local _0_
do
  local _0_ = {a = "b"}
  local a = _0_["a"]
  _0_ = a
end
t["a"] = _0_
```
The second `_0_` shadows the first and so `t.a` is always assigned `nil`, and it should have the value `"b"`.

technomancy:
The root cause here is that when you compile a `let`, we were compiling the binding vector first, then compiling the body. However, compiling the body requires doing a gensym for the body's return value, and that gensym is emitted in the Lua output *before* the binding vector's values. This causes the possibility for a conflict.

If the outer scope's gensym is generated first, then the binding vector's gensym can detect the collision and generate a non-conflicting local. So that's what's happening in this patch; we generate `preSyms` inside the `let` special and pass it thru to the code in `doImpl` which compiles the body.

This fixes #245.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -1378,7 +1378,7 @@ local function checkUnused(scope, ast)
 end
 
 -- Implements a do statement, starting at the 'start' element. By default, start is 2.
-local function doImpl(ast, scope, parent, opts, start, chunk, subScope)
+local function doImpl(ast, scope, parent, opts, start, chunk, subScope, preSyms)
     start = start or 2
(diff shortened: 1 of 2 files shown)


technomancy:
Thanks!

PR #255, opened 2020-05-02
Preserve source data on quoted forms

### LINKED ISSUE #253: Quote does not preserve source location data
When symbols and lists come from quoting, their location data is erased, leading to less-helpful compiler messages:

```
(macro abc [x]  `(let [y 2] ,x))

Compile error in unknown:?
  macro tried to bind y without gensym
```

technomancy:
This fixes #253 for symbols, autogensyms, and tables inside quote by generating Lua that sets the source metadata on the values emitted.

It does not yet fix the problem for quoted lists, because of how lists are generated. Quoting a symbol results in a call to the function `sym` which takes three args, the name, the scope, and a table for source details. So splicing source details in is easy. Quoting a list results in a call to `list`, a function whose only arguments are the elements of the list itself.

**Edit**: fixed it for lists too.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -557,7 +557,7 @@ local function assertCompile(condition, msg, ast)
         local line = m and m.line or ast.line or "?"
         -- if we use regular `assert' we can't provide the `level' argument of 0
         error(string.format("Compile error in '%s' %s:%s: %s",
-                            isSym(ast[1]) and ast[1][1] or ast[1] or '()',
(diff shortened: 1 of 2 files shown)


technomancy:
OK, turns out the IIFE for this wasn't actually *that* bad.

For a language that lacks `doto`, anyway.

```
local s = "(function(l) l.filename, l.line = %s, %s return l end)(list(%s))"
return (s):format(filename, form.line or "nil", mixedConcat(mapped, ", "))
```

PR #256, opened 2020-05-03
rename limit-values/limit-args to pick-values/pick-args

jaawerth:
Renamed `limit-values`/`limit-args` to `pick-values`/`pick-args` and updated the changelog and reference.

@technomancy:, the examples in the reference are significantly different from the ones you commented on (with the redundant `(select :# ...)`) in an earlier version of this branch; I realized the other examples, with `[(pick-values 3 :a :b)] ; => [:a :b nil]` were incorrect, since sequences do nothing with trailing nils.

## example when n > # values
I added an example of this to `pick-values` in the reference, to provide a heads up on possible bugs arising from `n` being greater than the number of values passed. I didn't add the same note to `pick-args`, but I did make sure one of the examples illustrated this.

## docstring formatting

I also tweaked the formatting of the docstrings so their output in the repl appears like:

```
>> (doc pick-values)
(pick-values n ...)
  Like the `values` special, but emits exactly n values.
  For example,
  	(pick-values 2 ...)
  expands to
  	(let [(_0_ _1_) (values ...)] (values _0_ _1_))

>> (doc pick-args)
(pick-args n f)
  Creates a function of arity n that applies its arguments to f.
  For example,
  	(pick-args 2 func)
  expands to
  	(fn [_0_ _1_] (func _0_ _1_))
```

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -2847,7 +2847,7 @@ expands to\n\t(fn [_0_ _1_] (func _0_ _1_))"
                  `(fn ,bindings (,f ,(unpack bindings)))))
  :pick-values (fn [n ...]
                  "Like the `values` special, but emits exactly n values.\nFor example,
-\t(pick-values 2 ...)\nexpands to\n\t(let [(_0_ _1_) (values ...)] (values _0_ _1_))"
+\t(pick-values 2 ...)\nexpands to\n\t(let [(_0_ _1_) ...] (values _0_ _1_))"
                  (assert (and (= :number (type n)) (>= n 0) (= n (math.floor n)))
                          "Expected n to be an integer >= 0")
                  (let [let-syms   (list)
--- a/fennel.lua
+++ b/fennel.lua
@@ -2836,15 +2836,18 @@ Same as ->> except will short-circuit with nil when it encounters a nil value."
             (let [body (list f ...)]
               (table.insert body _VARARG)
               `(fn [,_VARARG] ,body)))
- :limit-args (fn [n f]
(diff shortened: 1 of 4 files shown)


fennel.lua
```diff
@@ -2836,15 +2836,18 @@ Same as ->> except will short-circuit with nil when it encounters a nil value."
...
                (assert (and (= (type n) :number) (= n (math.floor n)) (>= n 0))
                  "Expected n to be an integer literal >= 0.")
                (let [bindings []]
                  (for [i 1 n] (tset bindings i (gensym)))
                  `(fn ,bindings (,f ,(unpack bindings)))))
- :limit-values (fn [n ...]
-                 "Limits values to the first n items. Useful for restricting multiple returns."
+ :pick-values (fn [n ...]
+                 "Like the `values` special, but emits exactly n values.\nFor example,
+\t(pick-values 2 ...)\nexpands to\n\t(let [(_0_ _1_) (values ...)] (values _0_ _1_))"
```
technomancy, line 2850: One question I'm actually not sure about: is the first `values` here necessary?

It seems to me that `(values ...)` should be equivalent to just `...` but I'm not sure whether my intuition on that is correct.

jaawerth: Good call! It only adds the first `(values ,...)` to the expansion if the macro has received more than one literal as input, to get around compile errors from `(pick-values 2 :a :b :c)` due to bad generated input to `let`. I'll update it in a moment.

technomancy:
Nice; thanks!

PR #258, opened 2020-05-03
Macro expander

### LINKED ISSUE #183: macroexpander
Another great feature we should add.

technomancy:
This implements a first-class macroexpander in the form of `macroexpand`, `macroexpand-1`, and `macroprint`, the latter of which is intended to be used interactively for debugging.

Bonus: also improved the docs for other macro-related reference materials.

Edit: renamed to `macroprint` to `macrodebug` and made it automatically fall back to `tostring` if it can't fennelview, with a note for how to make it able to find fennelview in the reference.

Fixes #183.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -520,6 +520,9 @@ local function makeScope(parent)
         specials = setmetatable({}, {
             __index = parent and parent.specials
         }),
+        macros = setmetatable({}, {
+            __index = parent and parent.macros
+        }),
         symmeta = setmetatable({}, {
             __index = parent and parent.symmeta
         }),
@@ -751,7 +754,7 @@ end
 local function checkBindingValid(symbol, scope, ast)
     -- Check if symbol will be over shadowed by special
     local name = symbol[1]
-    assertCompile(not scope.specials[name],
+    assertCompile(not scope.specials[name] and not scope.macros[name],
                   ("local %s was overshadowed by a special form or macro")
                       :format(name), ast)
     assertCompile(not isQuoted(symbol),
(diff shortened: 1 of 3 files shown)


jaawerth:
Nice! I'll take a look at this tomorrow.

As for `macroprint`, maybe `macrodebug`?

reference.md
```diff
@@ -801,6 +822,41 @@ more details about the macro interface.
...
+```fennel
+(macrodebug (-> abc
+                (+ 99)
+                (> 0)
+                (when (os.exit))))
+; -> (if (> (+ abc 99) 0) (do (os.exit)))
+```
+
+Call the `macrodebug` macro with a form and it will expand all the
+macros in that form and print out the resulting form. Note that the
```
jaawerth, line 847: This seems to imply that all macros in the form will be expanded, but this isn't quite accurate, since what really happens is that it repeatedly expands each macro form until it the top-level form is no longer a macro form, then prints it.

Even if we don't expand it fully, maybe we can come up with a terse way to hint that it won't expand nested subforms? Just so people don't expect a fully expanded/"atomic" form.

jaawerth:
Another question that occurred to me: if we're deprecating `get-scope`, but also exposing `macroexpand` and `macroexpand-1` (which take `scope` as an argument) in the compiler env, how will people be able to use them? Isn't `_SCOPE` the scope of the micro at definition time, not the scope where the macro is being used?

fennel.lua
```diff
@@ -2888,6 +2889,9 @@ that argument name begins with ?."
           (assert (sym? name) "expected symbol for macro name")
           (local args [...])
           `(macros { ,(tostring name) (fn ,name ,(unpack args))}))
+ :macrodebug (fn macrodebug [form]
+              `(print ,((or (pcall require :fennelvieww) tostring)
```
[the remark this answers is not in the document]
technomancy: Haha; I added that typo intentionally to test the fallback, and then I guess I left it in!

reference.md
technomancy: That's a great point; I hadn't even considered that crucial distinction. I've changed it to make it clearer that it only affects top-level calls.

technomancy:
Thanks for the feedback!

Most of the fixes were pretty simple to make, but the scoping problem is hairier. For the time being I've made it so `macroexpand` can only be called from macros to avoid the arbitrary scoping problem (`_SCOPE` isn't quite the right thing here since it won't always give us the nested scope if you call it when a macro is defined inside a function or something) but we can revisit this later if we have a strong use case for calling that function in other contexts.

jaawerth:
Thanks for your work on this, I'm excited to start using it!

PR #259, opened 2020-05-03
Implement stablepairs and use it when destructuring.

### LINKED ISSUE #257: destructuring k/v tables in Lua 5.3 gives non-deterministic order in compiler output
In Lua 5.2 and 5.3, table iteration order with `pairs` appears to be indeterminate. Because [destructuring k/v tables currently uses `pairs`](https://github.com/bakpakin/Fennel/blob/1ffff3f386e1c33362b9aeb33124f66a73b93d89/fennel.lua#L1288), the definition order in the compiled Lua output can change between runs.

The order is consistent in both PUC and LuaJIT versions of 5.1

## Test case

Running this in the shell, with `lua --v gives Lua 5.3.x or 5.2.x`:
```bash
(testcode='(let [{: a : b : c} {:a 1 :b 2 :c 3}] nil)';
for i in `seq 1 10`; do
  echo "$i\t" $(fennel --compile <(echo $testcode));
done)
```
```lua
1	local _0_ = {a = 1, b = 2, c = 3} local b = _0_["b"] local a = _0_["a"] local c = _0_["c"] return nil
2	local _0_ = {a = 1, b = 2, c = 3} local b = _0_["b"] local c = _0_["c"] local a = _0_["a"] return nil
3	local _0_ = {a = 1, b = 2, c = 3} local c = _0_["c"] local b = _0_["b"] local a = _0_["a"] return nil
4	local _0_ = {a = 1, b = 2, c = 3} local c = _0_["c"] local b = _0_["b"] local a = _0_["a"] return nil
5	local _0_ = {a = 1, b = 2, c = 3} local b = _0_["b"] local c = _0_["c"] local a = _0_["a"] return nil
6	local _0_ = {a = 1, b = 2, c = 3} local c = _0_["c"] local b = _0_["b"] local a = _0_["a"] return nil
7	local _0_ = {a = 1, b = 2, c = 3} local b = _0_["b"] local a = _0_["a"] local c = _0_["c"] return nil
8	local _0_ = {a = 1, b = 2, c = 3} local a = _0_["a"] local b = _0_["b"] local c = _0_["c"] return nil
9	local _0_ = {a = 1, b = 2, c = 3} local b = _0_["b"] local a = _0_["a"] local c = _0_["c"] return nil
10	local _0_ = {a = 1, b = 2, c = 3} local a = _0_["a"] local c = _0_["c"] local b = _0_["b"] return nil
```

## Possible fixes

1. The direct fix would be to have `destructure`/`destructure1` in the compiler first build and sort a table of keys , so the output is always consistent.
2. If we created an internal AST node type for k/v tables that keeps track of key order, we could prevent similar issues from cropping up if `pairs` is being used in the compiler, but it would be more work to change.

One possible approach to **2** is to create a `KV_TABLE_MT` that tracks key order, and sets `_pairs` on the metatable to use it. This might save us having to replace `pairs` usage throughout the computer: Lua 5.3 would use `_pairs` to control iteration, while older versions would compile differently from 5.3 but be internally consistent.

technomancy:
This is called from five places:

* twice inside `destructure`
* in `kvmap` which is used when emitting table literals
* destructuring that's done inside `each`
* `mixedConcat` which is responsible for printing quoted tables

My first implementation was very simple since I was only concerned
with destructuring, and destructuring never happens on large
tables. But when I realized it would be used for all table literals I
realized we wouldn't want to be so sloppy.

Fixes #257

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -30,6 +30,20 @@ local unpack = unpack or table.unpack
 -- Main Types and support functions
 --
 
+-- Like pairs, but gives consistent ordering every time. On 5.1, 5.2, and LuaJIT
+-- pairs is already stable, but on 5.3 every run gives different ordering.
+local function stablepairs(t)
+    local keys, succ = {}, {}
+    for k in pairs(t) do table.insert(keys, k) end
+    table.sort(keys, function(a, b) return tostring(a) < tostring(b) end)
+    for i,k in ipairs(keys) do succ[k] = keys[i+1] end
+    local function stablenext(tbl, idx)
+        if idx == nil then return keys[1], tbl[keys[1]] end
+        return succ[idx], tbl[succ[idx]]
+    end
+    return stablenext, t, nil
+end
+
 -- Map function f over sequential table t, removing values where f returns nil.
 -- Optionally takes a target table to insert the mapped values into.
 local function map(t, f, out)
@@ -48,7 +62,7 @@ end
 local function kvmap(t, f, out)
     out = out or {}
     if type(f) ~= "function" then local s = f f = function(x) return x[s] end end
-    for k,x in pairs(t) do
+    for k,x in stablepairs(t) do
         local korv, v = f(k, x)
         if korv and not v then table.insert(out, korv) end
         if korv and v then out[korv] = v end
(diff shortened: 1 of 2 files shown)


jaawerth:
Lookin' good

PR #260, opened 2020-05-05
make completion walk scope ancestors

jaawerth:
The recent merge on master stopped macros from showing up in REPL readline.lua completion because it was pointing at `scope.macros` instead of `scope.parent.macros`, and I realized this would be more reliable if the completer simply walked ancestors for each table of candidates.

- Adds `allPairs` function that walks ancestors, skipping duplicate keys
- remove internals.SPECIALS from repl and add internals.allPairs
- modify the completion items cap from 40 to 2000; this will stop it
from blowing up, and let the frontend (e.g. readline) handle how many to
display.

Could use some feedback on changing the matches limit to 2000 - we could just remove it entirely. Readline has a (configurable) internal limit after which it will ask you if you want to display all N matches, and we can probably just let the frontend handle that aspect, but a limit will stop runaway iteration if something weird happens (circular inheritance, some clobbering of globals, etc).

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -13,6 +13,7 @@
 * Set macros on scope.macros & add new `macroexpand` helper to expand macro forms during
 compilation rather than uplifting macros to specials. (#258)
 * Add `macrodebug` util macro for printing expanded macro forms in REPL. (#258)
+* Make completion walk scope ancestors for macros & specials (#260)
 
 ## 0.3.2 / 2020-01-14
 
--- a/fennel.lua
+++ b/fennel.lua
@@ -70,6 +70,26 @@ local function kvmap(t, f, out)
     return out
 end
 
+local function allPairs(t)
+    assert(type(t) == 'table', 'allPairs expects a table')
+    local seen = {}
+    local function allPairsNext(_, state)
+        local nextState, value = next(t, state)
+        if seen[nextState] then
+            return allPairsNext(nil, nextState)
+        elseif nextState then
+            seen[nextState] = true
+            return nextState, value
+        end
+        local meta = getmetatable(t)
+        if meta and meta.__index then
+            t = meta.__index
+            return allPairsNext(t)
+        end
+    end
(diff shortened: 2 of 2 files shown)


changelog.md
```diff
@@ -13,6 +13,7 @@
 * Set macros on scope.macros & add new `macroexpand` helper to expand macro forms during
 compilation rather than uplifting macros to specials. (#258)
 * Add `macrodebug` util macro for printing expanded macro forms in REPL. (#258)
+* Make completion walk scope ancestors for macros & specials (#260)
```
technomancy, line 16: This is fine, but most of the time I like keeping the changelog to just user-visible changes, and it seems pretty unlikely that anyone would have noticed this one. =)

fennel.lua
```diff
@@ -2456,11 +2476,11 @@ local replsource = [===[(local (fennel internals) ...)
   (let [matches []
         input-fragment (text:gsub ".*[%s)(]+" "")]
     (fn add-partials [input tbl prefix] ; add partial key matches in tbl
-      (each [k (pairs tbl)]
+      (each [k (internals.allPairs tbl)]
         (let [k (if (or (= tbl env) (= tbl env.___replLocals___))
                     (. scope.unmanglings k)
                     k)]
-          (when (and (< (# matches) 40)
+          (when (and (< (# matches) 2000) ; stop explosion on too many items
```
technomancy, line 2483: I think you're right that we should keep the limit just in case; if anyone needs more than 2000 it's likely they're doing something so specialized that they can do this part for themselves.

changelog.md
jaawerth: Good point. Feel free to strip it out when you cut the release!

jaawerth:
@technomancy While I did test it out in the repl, you raise a good point that it'd be a good idea to add some completion tests for macros, specials, and globals at some point, since right now every input it tests against relates to locals.

PR #262, opened 2020-05-07
WIP setup guide

### LINKED ISSUE #250: Setup guide
One missing piece in the documentation is it is now is a guide for setup and install of Fennel both for general-purpose use and for various contexts like embedding, love2d, etc. We should also cover readline installation and editor support.

We should take a look at various other languages' getting started guides too.

technomancy:
Still got a lot more to do on this, but we can use this PR to discuss.

Fixes #250.


technomancy:
Going to merge this as-is so I can set it up on fennel-lang.org; we can do another PR to master for vim/vs-code.

PR #263, opened 2020-05-08
edited setup.md

m455:


setup.md
```diff
@@ -4,166 +4,241 @@ This document will guide you through setting up Fennel on your
 computer. This document assumes you know how to download a Git
 repository and edit configuration files in a UNIX-like environment.
 
-## Requirements
+Fennel can be used in non-UNIX environments, but those environments
+will not be covered in this document.
+
+# Table of contents
```
technomancy, line 10: Can we skip the table of contents here and insert it using pandoc like we do with reference.md on fennel-lang.org? https://git.sr.ht/~technomancy/fennel-lang.org/commit/d1ac1f50d10f2fd09ac2d182c804e9b692a14374

IMO the table of contents is most useful when the hyperlinks work; when you're viewing it in your editor rather than your browser you don't have the benefit of being able to jump to a specific point.

setup.md
```diff
@@ -4,166 +4,241 @@ This document will guide you through setting up Fennel on your
...
+**Note**: You can try running `fennel --help` to confirm the
+installation succeeded.
 
 # Embedding Fennel
 
-Installing Fennel on your system will allow you to run scripts written
-in pure Fennel. However, for more complex situations it's common to
-embed Fennel code inside a larger application. You can do this in two
-ways: including the Fennel compiler or performing ahead-of-time compilation.
+Fennel can be embedded inside of Lua applications by including the
```
technomancy, line 101: Let's say `Fennel code can be embedded...` instead of `Fennel can be embedded`, because that makes it sound like you are putting Fennel (the compiler) in your program, but is not the case when you do AOT compilation.

setup.md
```diff
@@ -4,166 +4,241 @@ This document will guide you through setting up Fennel on your
...
+restricted, then compiling Fennel code to Lua during the build
+process, and including the Lua output in the application may be
+easier.
+
+## Embedding the Fennel compiler in a Lua application
+
+The Fennel compiler can be added to your code repository, and then
+loaded from Lua. This allows you to call functions defined in Fennel
+code from Lua, and functions defined in Lua from Fennel without
+slowdown.
```
technomancy, line 123: I realized that the bit about being able to call functions defined in Fennel from Lua and the other way around maybe shouldn't go in this section, because it's not specific to embedding the Fennel compiler; it is true in both cases.

setup.md
```diff
@@ -4,166 +4,241 @@ This document will guide you through setting up Fennel on your
...
-to perform the compilation. Add this to your `Makefile`:
+### To perform ahead-of-time compilation
+
+1. Run `touch Makefile`
+2. Add the following lines to the `Makefile`:
 
 ```
 %.lua: %.fnl fennel
-	./fennel --compile $< > $@
+    ./fennel --compile $< > $@
```
technomancy, line 158: Most of the time it's OK to replace tabs with spaces; however in this context (inside a Makefile) it will actually cause a syntax error. =( This is one of those things that's just really stupid about `make`, which is overall pretty good.

setup.md
```diff
@@ -4,166 +4,241 @@ This document will guide you through setting up Fennel on your
...
-It's recommended you include `fennel` itself in your repository so that
-you will always get consistent results rather than relying on whatever
-version of Fennel is installed on your machine at the time of building.
+3. Run `make` to perform the compilation
+
+**Note 1**: Ahead-of-time compilation is also useful if what you are
+working with requires optimal startup time. "Fennel compiles fast,
+but not as fast as not having to compile" -- jaawerth
+
+**Note 2**: It's recommended you include `fennel` itself in your
```
technomancy, line 167: Let's say "recommended you include the `fennel` script" here to be extra clear.

setup.md
m455: We sure can!

setup.md
m455: Good point! I'll do that!

setup.md
m455: Another option could be throwing that information into a "note", as it's additional information, but still related. Would you like to do this or remove it as you suggested? :D

setup.md
m455: Damnit, I'm not used to having tabs in my Markdown files. I totally didn't realize my editor did this. Thanks for catching this! haha, I will fix this!

setup.md
m455: Good call! I'll make that adjustment!

setup.md
m455: Ahhh I see what you meant now. How about I move it inside the main parent "# Embedding Fennel"?

setup.md
```diff
@@ -4,166 +4,241 @@ This document will guide you through setting up Fennel on your
...
-3. Change directories with `cd Fennel`
-4. Run `make fennel` to compile Fennel into a standalone script
-5. Copy or link `fennel` to a location on your shell's `$PATH` (eg
-   `~/bin` or `/usr/local/bin`)
+**Note 1**: `Step 2.` above will compile Fennel into a standalone script
+called `fennel`.
 
-## Using LuaRocks to download Fennel
+**Note 2**: If the `fennel` script exists inside your `$PATH`
+directory, you can run `fennel filename.fnl` to run a Fennel file.
```
Benaiah, line 82: "your `$PATH` directory" would be more accurately stated as "one of the directories on your "`$PATH`".

setup.md
```diff
@@ -4,166 +4,241 @@ This document will guide you through setting up Fennel on your
...
-* You want to use a version that hasn't been released yet
-* You want to contribute changes to Fennel
-* You have Git installed and don't want to bother with another system
+### To install Fennel
 
-To install:
+1. Run `cd Fennel`
+2. Run `make fennel`
+3. Copy or link `fennel` to a location on your shell's `$PATH`, such
+   as `~/bin` or `/usr/local/bin`
```
Benaiah, line 76: This seems to imply that `~/bin` is on the shell's `PATH` in most situations, which I don't believe is the case unless the user has already customized their `PATH` to include it. If the user has a customized `PATH`, they probably know where their custom bin directory is. I think we can remove `~/bin` here. `/usr/local/bin` makes sense to keep IMO tho.

setup.md
```diff
@@ -4,166 +4,241 @@ This document will guide you through setting up Fennel on your
...
+**Note 2**: If the `fennel` script exists inside your `$PATH`
+directory, you can run `fennel filename.fnl` to run a Fennel file.
+
+## Using LuaRocks to download and install Fennel
 
 [LuaRocks](https://luarocks.org/) contains a repository of Lua
-packages. It allows you to download and install packages and automates
-installation steps.
+software packages. LuaRocks is convenient because it automates the
+downloading, installation, and uninstallation of software packages.
```
Benaiah, line 88: "of Lua software packages" or just "of Lua software" maybe?

setup.md
```diff
@@ -4,166 +4,241 @@ This document will guide you through setting up Fennel on your
...
+easier.
+
+## Embedding the Fennel compiler in a Lua application
+
+The Fennel compiler can be added to your code repository, and then
+loaded from Lua. This allows you to call functions defined in Fennel
+code from Lua, and functions defined in Lua from Fennel without
+slowdown.
+
+### To embed the Fennel compiler in a Lua application
```
Benaiah, line 125: This header feels redundant with the header above it. Maybe demote it to a paragraph?

setup.md
```diff
@@ -4,166 +4,241 @@ This document will guide you through setting up Fennel on your
...
+  information on using external editors, instead of the built-in
+  TIC-80 editor.
 
 ## Using Fennel with LOVE
 
-LOVE has no built-in support for Fennel, so you need to bring it yourself.
-This [project skeleton repo](https://gitlab.com/alexjgriffith/min-love2d-fennel)
-for LOVE shows you how to do that, including a console-based REPL for
-debugging and reloading.
+LOVE has no built-in support for Fennel, so setup support yourself.
```
Benaiah, line 221: "so you'll need to set up support yourself" maybe?

setup.md
```diff
@@ -4,166 +4,241 @@ This document will guide you through setting up Fennel on your
...
+
+This [project skeleton for LOVE](https://gitlab.com/alexjgriffith/min-love2d-fennel)
+for LOVE shows you how to setup support for Fennel and how to setup a
```
Benaiah, line 224: This line reads "for LOVE for LOVE".

setup.md
m455: Thanks! I still have trouble conceptualizing `$PATH`!

setup.md
m455: Good call! This was me carrying out examples from the pre-edited version, and I wasn't sure if it was safe to remove.

setup.md
m455: I'll see what others decide on here, as I am not too familiar with LuaRocks in the first place :P

setup.md
m455: (Already chatted about this with you in `#fennel` on Freenode, so I'll resolve this soon)

setup.md
m455: Hahaha, thanks for catching this.

setup.md
m455: Oof. thanks

setup.md
m455: Chatted with @Benaiah about this, and have settled on removing it for now.

PR #264, opened 2020-05-08
Make fennel launcher script self-hosting.

### LINKED ISSUE #248: Create portable single-file Fennel CLI
Users might expect to be able to move the `fennel` script around as a single file and have it still work. While we may or may not want a single-file CLI script to be the default fennel script, it would be useful to have one in the repo for people who want to be able to just copy a single file into a directory on their `PATH`. It would also be useful to allow people to install the fennel compiler by downloading just one file - that would make `curl` or `wget` a reasonable option for installing fennel if you didn't want the whole repo.

technomancy:
This is a big change!

This moves the existing `fennel` script to `launcher.lua` and
introduces a new `launcher.fnl` script which is a pretty straight
port of the old version, except since it's in Fennel, we can use the
--require-as-include option in order to create a standalone script
that works as a single file.

This will make installation of releases much more straightforward
since now there is a single file.

There are two downsides:

1. There is risk that the port from Lua to Fennel introduced bugs.
2. Now when you do a fresh checkout, you must run `make` before you
   have a `fennel` script you can use. Every change you make to
   fennel.lua, fennelview.fnl, or fennelfriend.fnl will not get picked
   up until you run `make` again.

We could check a compiled `fennel` into the repository to help with
the first half of 2, but not the second part.

Fixes #248.


launcher.fnl
```diff
@@ -0,0 +1,172 @@
...
+          (do (readline.set_completion_append_character "")
+              (completer (text:sub from to)))
+          []))
+    (readline.set_complete_function repl-completer)
+    readline))
+
+;; TODO: generalize this as a plugin instead of hard-coding it
+;; we can't pcall this or we won't be able to use --require-as-include.
+(each [k v (pairs (require :fennelfriend))]
+  (tset options k v))
```
jaawerth, line 119: What if we pcalled a wrapper function instead?

```suggestion
;; TODO: generalize this as a plugin instead of hard-coding it
(match (pcall #(pairs (require :fennelfriend)))
  (true step param) (each [k v (values step param)]
                      (tset options k v)))
```

launcher.fnl
```diff
@@ -0,0 +1,172 @@
...
+      (tset opts k v))
+    (table.insert (or package.loaders package.searchers)
+                  (fennel.make_searcher opts))))
+
+(fn try-readline [ok readline]
+  (when ok
+    (when readline.set_readline_name
+      (readline.set_readline_name "fennel"))
+    (readline.set_options {:keeplines 1000 :histfile ""})
+    (fn opts.readChunk [parser-state]
```
jaawerth, line 101: Currently fails due to `opts` not being defined:

```suggestion
    (fn options.readChunk [parser-state]
```

launcher.fnl
```diff
@@ -0,0 +1,172 @@
...
+      (let [prompt (if (< 0 parser-state.stackSize) ".. " ">> ")
+            str (readline.readline prompt)]
+        (if str (.. str "\n"))))
+    (var completer nil)
+    (fn opts.registerCompleter [repl-completer]
```
jaawerth, line 106:
```suggestion
    (fn options.registerCompleter [repl-completer]
```

jaawerth:
One other thought regarding args parsing - maybe we could cook up a macro to cleanly define the flag options along with their aliases so it's easier to see what's being defined flags-wise at a quick glance? Not for this PR necessarily, that could always be added later.

Since we're now doing codegen on the CLI and there are some moving parts, we may want to consider adding a `tests/cli.fnl` that run various `./fennel` commands with readline on/off via `os.execute` just so we can more easily catch regressions. Again, could probably be added later.

launcher.fnl
technomancy: I'm on the fence about this... on the one hand, it's more resilient, but on the other hand, maybe the only reason we were using pcall in the first place is because you'd copy the old `fennel` script out of the repo and leave out the other files because we didn't have a standalone script. If we have a standalone script, what's the rationale for wanting to use just the launcher without the other goodies?

technomancy:
Great points all! Will see about getting those addressed.

launcher.fnl
jaawerth: That's a good point. The only scenario I can see it coming up is if someone does their own compilation and it somehow doesn't find fennelfriend, and ends up using the require fallback. It'd probably be better to get instant feedback on a failed `include` anyway, to catch any issues with fennelfriend.

I mostly just added it because the comment about not being able to `pcall` was still there, so just removing that comment line sounds good to me as well.

launcher.fnl
technomancy: How ... does this even compile???

jaawerth:
Almost forgot! If we're not going to check in the built `fennel`, we should probably `.gitignore` just to avoid accidental check-ins.

launcher.fnl
technomancy: Yeah, this is a bug in the compiler; it fails to catch unknown globals in this case.

technomancy:
I started to write a macro for handling the args, but the indentation is much worse when it's not using a built-in form and it just started to bug me.

I think I got the rest of the points addressed tho; thanks for taking a look! I did find #267 as a result of the `opts` issue.

PR #265, opened 2020-05-08
Setup guide edited 2

m455:


technomancy:
This is great; thanks! We'll add the rest (vim, vs code, and readline) in the other PR before merging. Nice work.

PR #266, opened 2020-05-08
Rename bitwise operators to match luajit names.

### LINKED ISSUE #261: rename new bitwise operators?
@bakpakin added Lua 5.3 bitwise operators in #251 (for issue #238), currently named `<<`, `>>`, `&`, `|`, `xor`, and `bnot`. For consistency, I'd like to suggest we update the names:

## New names
- `<<` to `lshift`
- `>>` to `rshift`
- `&` to `band`
- `|` to `bor`
- `xor` to `bxor`
- `bnot` stays the same

This is consistent, leaves symbols like `&` open for us to use in future, and matches the names used by LuaJit's `bit.*` extension library. If we add an optional module that sets fallback macros for earlier Lua versions, those same macros can be overridden to e.g. dispatch to the `bit.*` library when it's available.

## Downside

One possible downside to mirroring the `bit.*` names is that it would conflict if people want to follow the not-uncommon pattern of `(local {: lshift : rshift : bxor} bit)` in LuaJIT.

However, if not the above, adding a common prefix (even if it's just `b` or `bo` for "binary operator") would help keep cognitive debt to a minimum by providing a visual cue that these specials only work in Lua 5.3. Open to suggestions!

## Docstrings
It would also be a good idea to ensure the docstrings for all of them make it clear they only work in Lua 5.3.

technomancy:
Also add docstrings for them.

https://bitop.luajit.org/api.html

Fixes #261

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -2,7 +2,9 @@
 
 ## ??? / ???
 
-* Add bit-wise operators `>>`, `<<`, `|`, `&`, `bnot`, and `xor`
+* Ensure deterministic compiler output (#257)
+* Fix an issue where quoted forms did not have filename/line source data (#253)
+* Add bit-wise operators `rshift`, `lshift`, `bor`, `band`, `bnot`, and `bxor`
 * Friendlier compiler/parse error messages with suggestions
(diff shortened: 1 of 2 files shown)


jaawerth:
Nice!

Issue #236, opened 2020-04-08
readline enhancements


jaawerth:
Planned/proposed enhancement to the CLI REPL's readline support (some in progress, others not). Feedback and additional suggestions welcome!

## In progress
- [x] __Actually document the REPL's readline support and tab completion__ via the optional readline.lua dep so people know it's there and what it can do!
- [x] __Set the application name in readline for conditional `.inputrc` directives,__ e.g.
  ```inputrc
  $if fennel
    set enable-bracketed-paste On
    set blink-matching-paren On
  $endif
  ```
  This is waiting on a [patch I submitted to readline.lua](https://gitlab.com/peterbillam/pjb_lua/-/merge_requests/1/diffs) supporting setting the libreadline application name.
- [x] __Save the REPL history if `histfile` is non-empty in the readline options__
  If we add a couple new lines to the CLI, people can simply do the following in `.fennelrc` to have a persistent rolling history:
  ```clojure
  (when package.loaded.readline
    (package.loaded.readline.set_options {:histfile "~/.fennel_repl_history"}))
  ```

## Proposed/not started
- [x] __When readline.lua fails to load, add a message to the CLI greeting msg__
  about installing readline.lua for enhanced cli support, so new users know tab completion, history, etc is there via an optional dep.

bbuccianti:
> Actually document the REPL's readline support and tab completion via the optional readline.lua dep so people know it's there and what it can do!

Someone has taken this?

I can make an attempt if nobody is doing it.

technomancy:
@jaawerth did you have this already in progress, or is it in a good place to hand off?

jaawerth:
Yep, I've got something mostly written and should have something PR'd to the `setup-guide` branch this evening. That said, @bbuccianti, if you see anything you want to add or change to it, definitely feel free. I'll tag you when I submit the PR so you can take a gander!

technomancy:
I think this is landed on master but of course there's always room for further improvement!

PR #268, opened 2020-05-09
Initial pass on adding readline to setup guide

jaawerth:
This is likely more verbose than it needs to be, but I figured I would err on the side of too much information to make sure all of the most relevant info is included.

Feel free to suggest edits here if you'd like me to make them, or you can simply merge this into the `setup-guide` branch and make revisions from there.

Addresses documentation todo in #236

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -269,8 +269,124 @@ TODO: Explain one of the following:
 * What significance does this have to the user?
 * What's in it for the user? WIIFM (What's in it for me? (for the user))
 
-### To add readline support to Fennel
+The command-line REPL that comes with the `fennel` works out of the box, but
+the built-in line-reader is very limited in user experience. Adding
+[GNU Readline](https://tiswww.case.edu/php/chet/readline/rltop.html) support
+enables such user-friendly features as
+- Tab completion on the REPL that can complete on all locals, macros, and special forms
+- A rolling history buffer, which can be navigated, searched (`ctrl+r`), and optionally
+persisted to disk so you can search input from previous REPL sessions.
+- emacs (default) or vi emulation via readline's custom support for better line
+navigation
+- Optional use of additional readline features in `~/.inputrc`, such as blinking
+on matched parentheses or color color output
 
-1. TODO
-2. TODO
-3. TODO
+### Requirements for readline support
+
+All you need to enable readline support is:
+- GNU Readline installed on your system (you may already have it!)
+- [readline.lua](https://pjb.com.au/comp/lua/readline.html) lua bindings to libreadline
+
+The Fennel CLI REPL will automatically load and use the readline bindings when
+it can resolve the readline module, so that's all you need to get started.
+
+### Installing readline.lua
+For the official support on getting readline.lua, see the
+[official docs](https://pjb.com.au/comp/lua/readline.html#installation).
+
+The easiest way to get readline.lua is to install it with Luarocks, which
+will fetch the package and automatically compile the native bindings for you.
+
+#### Installing readline with LuaRocks
+
+```bash
+# to install globally on the system (requires admin privileges or sudo)
+$ luarocks install readline
+
+# to install to the user's local tree
+$ luarocks install --local readline
+
+# install for Lua 5.1 (including LuaJIT)
+$ luarocks install --lua-version=5.1 readline
+```
+
+Because the readline Lua module contains native bindings to libreadline, be sure
+it's installed for the Lua version you intend to use.
+
+If you've installed with the `--local` flag, you may need to ensure your `package.path`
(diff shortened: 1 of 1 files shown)


jaawerth:
@bbuccianti, since you expressed interest in taking a pass on this today, feel free to PR any revisions to this you'd like, either by suggestion them now or via another PR once it's merged. I wouldn't be surprised if I documented things in an overly complicated manner, so I'm curious what you had in mind!

setup.md
```diff
@@ -269,8 +269,124 @@ TODO: Explain one of the following:
...
+The command-line REPL that comes with the `fennel` works out of the box, but
+the built-in line-reader is very limited in user experience. Adding
+[GNU Readline](https://tiswww.case.edu/php/chet/readline/rltop.html) support
+enables such user-friendly features as
+- Tab completion on the REPL that can complete on all locals, macros, and special forms
+- A rolling history buffer, which can be navigated, searched (`ctrl+r`), and optionally
+persisted to disk so you can search input from previous REPL sessions.
+- emacs (default) or vi emulation via readline's custom support for better line
+navigation
+- Optional use of additional readline features in `~/.inputrc`, such as blinking
```
technomancy, line 281: Let's link to a reference for `.inputrc` here; when I read this I thought to myself "I wonder what the options for this are; maybe I'm missing out". Edit: I see now it's covered below; maybe mention "see below" here.

setup.md
```diff
@@ -269,8 +269,124 @@ TODO: Explain one of the following:
...
+
+Because the readline Lua module contains native bindings to libreadline, be sure
+it's installed for the Lua version you intend to use.
+
+If you've installed with the `--local` flag, you may need to ensure your `package.path`
+and `package.cpath` contain its location so it can be resolved by `require`.
+
+**TODO: Add and link to section on configuring environment to detect local rocks**
+
+#### Manual installation
```
technomancy, line 321: TBH I think for a manual installation we should just refer to the upstream docs. I don't think it belongs in a general setup guide.

setup.md
```diff
@@ -269,8 +269,124 @@ TODO: Explain one of the following:
...
+```fennel
+; persist repl history
+(match package.loaded.readline
+  rl   (rl.set_options {:histfile  "~/.fennel_history" ; default:"" (don't save)
+                        :keeplines 1000}))             ; default:1000
+```
+
+#### Configuring readline in `~/.inputrc`
+
+[documentation on the readline init file](https://www.gnu.org/software/bash/manual/html_node/Readline-Init-File.html)
```
technomancy, line 366: Let's make this a complete sentence.

setup.md
```diff
@@ -269,8 +269,124 @@ TODO: Explain one of the following:
...
+
+```inputrc
+$if fennel
+  set enable-bracketed-paste on
+  set blink-matching-paren on
+  set show-all-if-ambiguous on
+$endif
+```
+
+**NOTE:** The conditional directive `$if fennel` used above to set fennel-only
```
technomancy, line 388: I think it's OK to assume the latest versions of each of these in the setup guide, since you won't get an older version if you follow the instructions in this file?

setup.md
```diff
@@ -269,8 +269,124 @@ TODO: Explain one of the following:
...
-1. TODO
-2. TODO
-3. TODO
+### Requirements for readline support
+
+All you need to enable readline support is:
+- GNU Readline installed on your system (you may already have it!)
+- [readline.lua](https://pjb.com.au/comp/lua/readline.html) lua bindings to libreadline
+
+The Fennel CLI REPL will automatically load and use the readline bindings when
```
technomancy, line 290: Let's call it "the stock Fennel REPL" or "the `fennel` command's REPL" here instead of "The Fennel CLI REPL".

technomancy:
I'll just make these changes on my branch.

setup.md
jaawerth: Oh man that's so much better. I keep saying Fennel CLI REPL to keep it distinct from an nrepl or editor repl, but it feels awkward every time!

setup.md
jaawerth: ... not sure how i let that happen :clown_face:

setup.md
jaawerth: True! And feel free to edit it at will, but I'd like to mention the requirements for `.inputrc` to work *somewhere* just it's written down somewhere people can find and we can link to in the event people take note of their `.inputrc` doing nothing.

This is the first time we documented the readline stuff, so it's possible very few people are currently using it, so I'm guessing there will be a lot of people who start using it. We might want to make mention of it in the changelog, just so people not using emacs know they can get some bells and whistles?

PR #269, opened 2020-05-09
[experimental] import-macros

### LINKED ISSUE #180: import-macros: alternative to require-macros with finer control
I think we need a way to bring in macros in a more controlled manner - since require directly sets values on the current macro "namespace" (not sure what we'd call it - scope?), there should be a way to only pull in the macros we want from a target module to avoid clobbering. I would even suggest this be the default, if possible.

## Considerations
1. It might be nice to rename `require-macros` to `import-macros` since, unlike `require`, it sets things in namespace/macro scope rather than returning a value to be used.
2. It should ideally default to targeting a specific list of macros to bring in from a file, with an option to bring in all of them.

## API thoughts
Off the top of my head, something similar to [clojure's `require`](https://clojuredocs.org/clojure.core/require) would work, modified to account for the different package systems. One difference - unless we want to allow grouping/namespacing of macros onto a fake table, we could omit `:refer` and make it implied.

`(import-macros <path.to.module1> [<imports>...] ...)`

```clojure
(import-macros
  :shims-or-wat :all  ; bring in everything
  :mypkg.utils [moe larry [curly :as shemp]]) ; alias curly to shemp when importing
```

Thoughts? It's a breaking API change to default to named macro imports and require opting into full clobbering, but I think it's the sanest default behavior.

As for the name change, just a thought - I could see that not being worth the confusion, but `import` makes sense to me as a way to keep the semantics distinct.

jaawerth:
Resolves #180 

This introduces the long-awaited `import-macros` as a new experimental feature. It allows pulling macros in from an external macro module without clobbering the scope with all exported macros, instead either setting them on a "namespace" or using k/v destructuring to explicitly import specific macros, optionally aliased to a different name.

A few notes:
- It's implemented as a macro, and required de-deprecating `(get-option)` to do so.
- Since destructuring is tied to Lua emission at the moment, this works instead by doing a one-level destructure of a k/v table. This shouldn't be too big a deal; it means you can't export/import sub-tables of macros, but that may not be something we want to allow anyway.
- This also modifies `macroexpand` to try and follow multisyms to find a macro in a "namespace", which is a newly-introduced term for a compiler-time table of macros on `scope.macros`.

## Usage Example
Considering macros modules
- `macros/nums.fnl` that returns `{: inc : dec}` : append }`
- `macros/table.fnl` that returns `{: append : tset-all}`

```clojure
(import-macros num                       :macros.nums
               {: tset-all}              :macros.table)
(local t (tset-all {} :a 1 :b 2))
(tset-all t :c 3 :d 5)
(set t.d (nums.dec t.d))
t ; => {:a 1 :b 2 :c 3 :d 4}
```
                         
## Known issues/edge cases

One reason I want to mark this is experimental is that there may be issues with using it. I've only identified one big one so far, but we'll want to test it thoroughly before we remove the "experimental" warning. This is also why we might consider a flag.

### Aliasing macros that are used by other macros 

The only issue I know some people may run into with this is in the event somebody imports a macro under a namespace, or under an aliased name, when another imported macro uses it.

For example, with macro module:

```clojure
(λ inc [n]
  (if (not (list? n)) `(+ ,n 1) `(let [num# ,n] (+ num# 1))))
(λ inc! [n-sym]
  (assert (sym? n-sym) "inc! expects a symbol targeting a variable")
  `(set ,n-sym (inc ,n-sym)))
``` 
Trying to import and use `inc!` as follows would result in an error due to `inc!` attempting to invoke a non-existent `inc!` macro instead of `num.inc!`:

```clojure
(import-macros num :macros.num)
(var i 0)
(num.inc! i)
```


jaawerth:
Here's an interesting side effect to the change I made to `macroexpand`: if you write a macro module that returns a nested object of key/macor-table pairs, you can avoid clobbering the scope even with `require-macros`:

screenshot
----


While that's neat, we may want to disallow it by asserting that every key on the exported table is a function, that or implement recursive k/v destructuring for `import-macros`. Thoughts?

fennel.lua
```diff
@@ -2950,6 +2955,35 @@ that argument name begins with ?."
...
+                    (if (sym? bindings)
+                        ;; import a whole set of macros to a symbol namespace
+                        (do (tset scope.macros (. bindings 1) {})
+                            (each [k v (pairs subscope.macros)]
+                              (tset (. scope.macros (. bindings 1)) k v)))
+
+                        ;; 1-level table destructuring for importing individual macros
+                        (table? bindings)
+                        (each [macro-name [import-key] (pairs bindings)]
+                          (assert (. subscope.macros macro-name)
```
technomancy, line 2982: I think you're right that we should assert that type=function here, not just non-nil.

fennel.lua
```diff
@@ -2660,13 +2671,7 @@ local function makeCompilerEnv(ast, scope, parent)
...
         ["sequence?"] = isSequence,
         ["varg?"] = isVarg,
-        ["get-scope"] = function()
-            if io and io.stderr then
-                io.stderr:
-                 write(("-- %s:%s get-scope is deprecated and will be removed\n"):
-                 format(ast.filename, ast.line))
-            end
-            return macroCurrentScope end,
+        ["get-scope"] = function() return macroCurrentScope end,
```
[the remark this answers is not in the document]
jaawerth: I think you might have this backwards?  `get-scope` is returning `macroCurrentScope`, which gets set (temporarily) to the scope at the point of expansion by `macroexpander`, so wouldn't that make it the current nested scope?

fennel.lua
```diff
@@ -1069,7 +1069,18 @@ end
 local macroCurrentScope = GLOBAL_SCOPE
 
 local function macroexpand(ast, scope, once)
+    local multiSymParts = isMultiSym(ast[1])
     local macro = isSym(ast[1]) and scope.macros[deref(ast[1])]
+    if not macro and multiSymParts then
+        local inNamespace
```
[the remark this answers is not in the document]
jaawerth: Works for me!

fennel.lua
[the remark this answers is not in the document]
technomancy: Yep; I was confusing it with definition-time scope.

fennel.lua
jaawerth: I can do that, but what about in `require-macros`? With the way this PR modifies `macroexpand` to check multisyms that resolve to a macro, it would still be possible to export nested macro objects unless we do the sanitizing in `require-macros`.

That doesn't preclude making the change here to check the binding form, just curious if you think I should disallow `{:foo {:bar (fn [] ...) :baz (fn [] ...) }}` in `require-macros` too.

technomancy: Oh, I guess I was just thinking that every leaf node should be a function, which we can't really solve here. I forgot that the single-level limitation applied only to destructuring, not to the overall shape of the module's table. Never mind!

jaawerth: Come to think of it, it wouldn't be very hard to make this tail-recursive, which would allow nesting. The question is: do we want to? It does potentially add complexity, but it would allow thinks like

- `macros/foo.fnl`
- `macros/bar.fnl`
- `macros/init.fnl` - returns `{:foo foo-macros :bar bar-macros}`

technomancy: No, I think this would give the impression that there is a relationship between filenames which doesn't actually hold. Different modules that resolve to files in the same directory do not actually interact with each other unless you make them, and we shouldn't introduce a connection that the module system on its own does not.

jaawerth:
Updated!

- I added a couple of named arguments, which will result in a nice `(doc import-macros)`
- Replaced references to "namespace" with "macro module"
- Added some suggestions to fennelfriend
- Modified `macroexpand` to verify the table returned from a macro module has only functions a properties
- Added a couple of `import-macros` failure tests, as well as ensured `addMacros` (via the `macros` special) is checking against non-function macros in the table

technomancy:
Brilliant; thanks!

PR #274, opened 2020-05-12
Add documentation in reference.md for import-macros.

technomancy:
I think this is the last missing piece for the 0.4.0 release!

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,7 +1,8 @@
 # Summary of user-visible changes
 
-## ??? / ???
+## 0.4.0 / ???
 
+* Add `import-macros` for more flexible macro module loading (#269)
 * Ensure deterministic compiler output (#257)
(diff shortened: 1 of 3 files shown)


changelog.md
```diff
@@ -2,6 +2,7 @@
 
 ## ??? / ???
 
+* Add `import-macros` for more flexible macro module loading (#180)
```
jaawerth, line 5: Maybe reference #269 here instead/in addition, since it's got a more concise write-up of the intent than the thread in #180?

jaawerth:
Looks good!

Was omitting **(Experimental)** in the reference for `import-macros` intentional? I'm guessing so, since the language for `require-macros` is also pushing `import-macros`.

technomancy:
Oops; no, omitting the experimental warning was an oversight; I've added it back in.

jaawerth:
:+1:

PR #275, opened 2020-05-12
fix tests to give proper exit code on failure

jaawerth:
*EDIT: Updated this to avoid the pitfalls of self-hosting when we may want to use the tests to see where something breaks.*

I noticed that test/init.lua was running tests on each module separately, and that `make test` was giving a 0 exit code even when there were failing tests.

While we could have tweaked it to keep a running tally on failures in each module, I took the opportunity to:
- Build and run a *single* test suite
- Ensure a proper non-zero exit code when tests fail
- Modify default test output to default to TAP format (human- and machine-readable Test Anything Protocol), which shows a minimal line for each test in the suite.
- Make the test output format configurable via FNL_TEST_OUTPUT env var, so it can be set to anything LuaUnit supports
- Clean up output of `make testall` to show only summaries, while `make test` defaults to outputting TAP.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,11 +3,12 @@ LUA ?= lua
 test: fennel
 	$(LUA) test/init.lua
 
+testall: export FNL_TEST_OUTPUT ?= text
 testall: fennel
-	lua5.1 test/init.lua
-	lua5.2 test/init.lua
-	lua5.3 test/init.lua
-	luajit test/init.lua
+	@echo -e 'Testing lua 5.1:'  ; lua5.1 test/init.lua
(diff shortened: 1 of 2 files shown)


jaawerth:
Belatedly realized this should be precompiled (or put back in Lua) so it can still be partially run in the event fennel gets really broken; that way we can at least see which tests fail. I'll see about fixing that tomororw.

jaawerth:
Ok, all updated. You can now override the format, so e.g. for TAP output with `make ci` you can do `FNL_TEST_OUTPUT=tap make ci`, or `FNL_TEST_OUTPUT=text make test` to get only a text summary.

technomancy:
Nice work; thanks!

PR #277, opened 2020-05-14
Fix macro args length calculation

### LINKED ISSUE #276: [BUG] 0.4.0 Regression in -?>> macro
-?>> macro may not apply all the args on the last function.  The -?> macro seems to work as expected though.

# Reproduction
Eval the following script
```clojure
(fn identity
  [x]
  x)

(fn join
  [sep list]
  (table.concat list sep))

(print "\nTHREAD"
      (hs.inspect
        (-?>> [:a :b]
              (join " "))))
```

## Expected

Output : "a b"

## Actual

An error is thrown
```
2020-05-13 21:48:44: *** ERROR: <file>:104: bad argument #1 to 'concat' (table expected, got nil)
stack traceback:
	[C]: in function 'table.concat'
```

## Notes

If you eval the following using identity as the last form, it works:

```
(fn identity
  [x]
  x)

(fn join
  [sep list]
  (table.concat list sep))

(print "\nTHREAD"
       (hs.inspect
        (-?>> [:a :b]
              (join " ")
              (identity))))
```

### Meta
- Discovered in https://github.com/agzam/spacehammer/issues/57

thatismatt:
The ast length can change as part of macro expansion, so this PR delays this calculation until after macro expansion.

Fixes #276

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -1126,9 +1126,9 @@ local function compile1(ast, scope, parent, opts)
     -- Compile the form
     if isList(ast) then
         -- Function call or special form
-        local len = #ast
-        assertCompile(len > 0, "expected a function, macro, or special to call", ast)
+        assertCompile(#ast > 0, "expected a function, macro, or special to call", ast)
         ast = macroexpand(ast, scope)
+        local len = ast and #ast or 0
         -- Test for special form
         local first = type(ast) == "table" and ast[1]
         if isSym(first) then -- Resolve symbol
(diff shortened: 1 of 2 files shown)


jaawerth:
Nicely spotted! Do you mind adding a quick test case to the threading macro cases within `test_macros` in `test/core.lua`?

Thanks!

thatismatt:
Good call, done.

jaawerth:
:+1:

PR #280, opened 2020-05-15
fix 0.4.0 macro expansion to primitives

### LINKED ISSUE #279: [0.4.0] macros can't return non-nil primitives
This is most likely related to #276; just a few more edge cases we missed after changing the compiler internals to use `macroexpand`. Should be a quick fix.

The following code works as expected in 0.3.2, but with the changes we made to how the macro system works internally for 0.4.0. Just a few things to iron out about how it handles primitives (non-lists/forms) returned from macros.

## Repro

```clojure
(macro str [] :hi)
(macro num [] 5)
(macro bool [] true)

(str) ;  => nil

(bool)
; => Compile error: ./fennel.lua:1072: attempt to index local 'ast' (a boolean value)

(num)
; => Compile error: ./fennel.lua:1072: attempt to index local 'ast' (a number value)
```

The usecase for doing anything resembling the above is relatively rare, but can sometimes be useful in certain codegen usecases.

jaawerth:
Fixes #279 

This tweaks `compile1` so that, if `macroexpand` returns a different AST than the one passed in (signifying a macro was expanded), it makes a recursive tail call to `compile1` so that it can handle each `ast` type anew instead of being stuck in the `isList(ast)` branch. This should simplify things so we don't have to special-case every possible macro expansion.

Because changes to `compile1` is at a relatively high risk of causing regressions, this could probably use some extra review just to make sure I didn't break anything else in the process.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -1123,21 +1123,16 @@ end
 local function compile1(ast, scope, parent, opts)
     opts = opts or {}
     local exprs = {}
-
+    -- expand any top-level macros before parsing and emitting Lua
+    ast = macroexpand(ast, scope)
     -- Compile the form
     if isList(ast) then -- Function call or special form
-        local len, oldAst = #ast, ast
         assertCompile(#ast > 0, "expected a function, macro, or special to call", ast)
-        ast = macroexpand(ast, scope)
-        -- if it's a macro, recur with the expanded ast
-        if (ast ~= oldAst) then return compile1(ast, scope, parent, opts) end
         -- Test for special form
-        local first = len > 0 and ast[1] or nil
-        if isSym(first) then first = first[1] end -- Resolve symbol
+        local len, first = #ast, ast[1]
         local multiSymParts = isMultiSym(first)
-        local special = scope.specials[first]
-        if special and isSym(ast[1]) then
-            -- Special form
+        local special = isSym(first) and scope.specials[deref(first)]
+        if special then -- Special form
             exprs = special(ast, scope, parent, opts) or expr('nil', 'literal')
             -- Be very accepting of strings or expression
             -- as well as lists or expressions
--- a/fennel.lua
+++ b/fennel.lua
@@ -1069,6 +1069,7 @@ end
 local macroCurrentScope = GLOBAL_SCOPE
 
 local function macroexpand(ast, scope, once)
+    if not isList(ast) then return ast end -- bail early if not a list form
     local multiSymParts = isMultiSym(ast[1])
     local macro = isSym(ast[1]) and scope.macros[deref(ast[1])]
     if not macro and multiSymParts then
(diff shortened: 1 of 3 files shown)


jaawerth:
There is *one* edge case this doesn't quite cover:

```clojure
(macro nope [] false)
(nope) ; => nil
```

In 0.3.2, the above expands correctly to `false`. With this PR in place it expands to `nil` instead of erroring, which is at least falsy, but not 100% correct. This prevents a compile error, however, and `(macro yes [] true)` returns correctly, so it gets us mostly there.

I'm guessing there's an implicit `falsy` check somewhere else in the code turning `false` into nil. I thought initially it might be `local first = len > 0 and ast[1] or nil`, but `first isn't being used anywhere that would cause this, so it's likely elsewhere...

jaawerth:
Ok, got the `false` -> `nil` edge case!

fennel.lua
```diff
@@ -1164,8 +1164,7 @@ local function compile1(ast, scope, parent, opts)
             end
             local compiled = compile1(newAST, scope, parent, opts)
             exprs = compiled
-        elseif(type(ast) == "table") then
```
technomancy, line 1167: What's the purpose of removing this check? Isn't the rest of this only sensible on a table?

fennel.lua
```diff
@@ -1069,6 +1069,7 @@ end
 local macroCurrentScope = GLOBAL_SCOPE
 
 local function macroexpand(ast, scope, once)
+    if not isList(ast) then return ast end -- bail early if not a list form
```
technomancy, line 1072: Rather than bailing early, shouldn't we just skip the recursion at the end when we expand to a non-list? Bonus: avoiding early returns will make it easier when we port it to self-hosting.

fennel.lua
```diff
@@ -1124,16 +1125,15 @@ local function compile1(ast, scope, parent, opts)
...
     -- Compile the form
-    if isList(ast) then
-        -- Function call or special form
+    if isList(ast) then -- Function call or special form
+        local len, oldAst = #ast, ast
         assertCompile(#ast > 0, "expected a function, macro, or special to call", ast)
         ast = macroexpand(ast, scope)
-        local len = ast and #ast or 0
+        -- if it's a macro, recur with the expanded ast
+        if (ast ~= oldAst) then return compile1(ast, scope, parent, opts) end
```
technomancy, line 1133: Shouldn't this conditional be specifically about whether `ast` is a list or not? If the new AST is different from the old AST, but it's still a list, then the recursion is unnecessary, right?

jaawerth: Good question - this was sort of my way of cutting the Gordian Knot ;-)

Looking at `compile1` as a whole, this check is occurring inside the `if isList(ast)` check at the very beginning. The macroexpansion is the only way for `ast` to be a non-table in this part of the code, at which point we'd basically want to start fresh branching on what kind of ast node it is. The cleanest way to do this struck mas a just invoking `compile1` again rather than having to restructure the `if/else` branches.

fennel.lua
jaawerth: Since the only way for this condition to be true in this part of the code (see the comment about recursion after `macroexpand`) because of the parent `if isList(ast)`, this can safely go back to an `else` since we're still operating under the parent condition being true.

fennel.lua
jaawerth: That occurred to me as well, but bailing early also makes it a bit more resilient. You can call `macroexpand` on an AST node without it blowing up in your face. If we want to give it stricter behavior, we could certainly move the `isList` check to the bottom. My update already skips recursion for a falsy AST node, so it would be a minor change, I just thought leaving the early return in makes this a bit safer to use.

technomancy: Hm; well if the `macroexpand` function can accept any AST form, not just a list, then why not call it from the top level of `compile1` instead of only inside the `isList` branch? Then the recursion would no longer be necessary, right?

jaawerth: Ha! Fair point, I'm kinda double-dipping in the defensive coding and  `compile1` needs all the readability it can get. I'll reorganize things a bit.

jaawerth:
@technomancy good call on taking the time to pull macro expansion to the top of `compile1`, it ended up being much cleaner that way. I took the time to clean up some of the `first` symbol-juggling logic during special detection, so please give the update a close look; I don't think I broke anything else in the process (and the tests still pass) but it could use a sanity check!

technomancy:
This is great; thanks! I think that really tightens up the logic.

Issue #278, opened 2020-05-14
[0.4.0] self-contained cli not quite self-contained


jaawerth:
Noticed that having another fennel version on my `package.path` was causing that to be loaded by the supposed-to-be-self-contained launcher script, overriding the version that should be preloaded.

I suspect this is caused by an ordering issue with `--require-as-include`; we may just need to tweak the order preloads are generated at the top of the file when there are multiple `include`s.

jaawerth:
I've narrowed this down to the root cause; because the `include` special does the preloading as follows when the include target is a Fennel module:

```lua
local forms = list(sym('or'), targetForm, list(sym('fn'), sequence()))
local p = parser(stringStream(s), path)
for _, val in p do table.insert(forms[3], val) end
compile1(forms, subscope, subChunk, subopts)
```

The above use of `list(sym('fn'), ..)` results in pointlessly setting metadata (in this case an empty arglist) on the generated loader function. This generates a `require("fennel")` that bypasses `--require-as-include` due to it being directly emitted by the `fn` special.

The three approaches that come to mind for fixing this are:

1. Modify `include` to not compile the loader set on `package.preload`, but to instead emit it directly as a function, and compile the source module to a string and simply append it to the body.
2. Modify the `fn` special to override the overall `useMetadata` setting.
3. Modify `fn` to skip setting metadata if there's no docstring and the arglist is empty.

technomancy:
I've been thinking of a different approach for `--require-as-include` anyway. The current one is limited only to direct `require` calls, which makes self-contained builds of limited use since they quit working as soon as one lua module loads another one.

What if `--require-as-include` functioned by introducing a `package.searcher` entry? It could load the module in question, which would cause the searcher to be re-invoked for each new transitive `require`.

The downside of course would be that people might not expect compilation to cause code to run, but I don't really see any other way of getting `--require-as-include` to work reliably.

jaawerth:
I'd be wary of doing this for 0.4.1. While writing a module with top-level side-effecting code isn't a great idea and should be pretty rare, I suspect this would make the use of `--require-as-include` more brittle than it already is, considering scenarios like:

- a module contains env-dependent conditional require calls based on platform, env vars, optional deps, etc; this could lead to different bundles based on conditionals at compile time for checks that were meant to happen during runtime
- a module is performing system side effects like writing to a file, etc (should be pretty rare, but it's not impossible someone may want to require a file that just performs an action and wasn't written as a module)
[...]

PR #281, opened 2020-05-16
Add --compile-binary command to make standalone binaries

technomancy:
This is still super experimental!!!

Edit: renamed it from `--compile-static` to `--compile-binary` to reflect that it can be used with or without static linking.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,9 +2,9 @@
 
 # Created by https://www.gitignore.io/api/lua
 
-# from luastatic
-/fennel-static
-/fnl.lua.c
+# from static compilation
+/*_binary.c
+/fennel-bin
 
 ### Lua ###
 # Compiled Lua sources
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,8 @@ luacheck:
 
 count:
 	cloc fennel.lua
-	cloc --force-lang=lisp fennelview.fnl fennelfriend.fnl launcher.fnl
+	cloc --force-lang=lisp fennelview.fnl fennelfriend.fnl fennelbinary.fnl \
+		launcher.fnl
 
 # For the time being, avoid chicken/egg situation thru the old Lua launcher.
 LAUNCHER=./old_launcher.lua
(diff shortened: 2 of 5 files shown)


technomancy:
I think this is ready for review. One question: does this merit its own module, or should it just be part of `launcher.fnl`? I'm leaning a bit towards keeping it separate to emphasize the fact that it's extremely untested and likely to be volatile, but maybe that's unnecessary.

jaawerth:
All else being equal, all else being equal I'd opt to keep it separate for *now*, only so we can play around with it and see if we want to change how it's configured a bit; even with a warning, it we add it as a supported flag to the standalone, some people may end up getting used to it being there so any future changes to the flags/env vars would be a breaking change, experimental or not. 0.4.1 is also primarily a bugfix release, and even though this is purely additive it might be nice if we stuck with stabilizing the core of 0.4.1 rather than risking future complications?

On the other hand, it IS purely additive so as long as we triple-check for side effects in the standalone build, so if you really want it in there I don't feel strongly enough about it to fight you on it :wink:.

As for review, if we *do* end up including it as part of `launcher.fnl` I'd just ask that we wait on merging this until a fix for #278 is merged, so we can test the standalone launcher for quirks free of complications with and without `fennelstatic`.

jaawerth:
Also, this is super cool!

technomancy:
> as long as we triple-check for side effects in the standalone build

This patch touches 4 files: `.gitignore`, `Makefile`, the new `fennelbinary` module, and `launcher.fnl`; so `fennel.lua` is unaffected. The only changes to `launcher.fnl` are to handle `--compile-binary` at the very end, and the help text.

>  if we do end up including it as part of launcher.fnl I'd just ask that we wait on a full review/merge of this until a fix for #278 is merged

Yep, good call.

technomancy:
I tidied up the C shim bits into one longer string and replaced the Lua loader shim with a Fennel macro to reduce the amount of code-in-a-string we rely upon. This is a very strange piece of code because we don't want to run the macro, we just want to take the compiled Lua as a string that comes from the macroexpansion and splice it into the C code after hex-encoding it.

Currently there's no way to do this, so I added a flag to `macrodebug` which returns the expanded form string instead of printing it.

I was also considering keeping the C shim bits in a separate file and loading it in at compile time but I haven't done that yet.

PR #283, opened 2020-05-18
Emit raw Lua function for package.preload of Fennel modules in include.

### LINKED ISSUE #278: [0.4.0] self-contained cli not quite self-contained

technomancy:
The previous implementation wrapped this in a Fennel function by constructing a list with `fn`. That would normally be fine, but `fn` causes metadata to be emitted. Metadata is stored by requiring the `fennel` module and calling `fennel.metadata:setall(...)`. Triggering this require during the load of the code causes the version of Fennel supplied by `include` to get ignored, leading to a situation where the "self-contained" build of Fennel ends up accidentally using the version of Fennel installed by Luarocks or some other package manager that sets `LUA_PATH`.

Fixes #278

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -2749,39 +2749,39 @@ SPECIALS['include'] = function(ast, scope, parent, opts)
 
     -- Read source
     local f = io.open(path)
-    local s = f:read('*all')
+    local s = f:read('*all'):gsub('[\r\n]*$', '')
     f:close()
 
     -- splice in source and memoize it in compiler AND package.preload
     -- so we can include it again without duplication, even in runtime
-    local target = 'package.preload["' .. mod .. '"]'
     local ret = expr('require("' .. mod .. '")', 'statement')
+    local target = ('package.preload[%q]'):format(mod)
+    local preloadStr = target .. ' = ' .. target .. ' or function()'
 
-    local subChunk, tempChunk = {}, {}
-    emit(tempChunk, subChunk, ast)
-    -- if lua, simply emit the setting of package.preload
-    if not isFennel then
-        emit(tempChunk, target .. ' = ' .. target .. ' or function()\n' .. s .. '\nend', ast)
-    end
+    local tempChunk, subChunk = {}, {}
+    emit(tempChunk, preloadStr, ast)
+    emit(tempChunk, subChunk)
+    emit(tempChunk, 'end', ast)
     -- Splice tempChunk to begining of rootChunk
-    for i, v in ipairs(tempChunk) do
-        table.insert(rootChunk, i, v)
-    end
+    for i, v in ipairs(tempChunk) do table.insert(rootChunk, i, v) end
 
     -- For fnl source, compile subChunk AFTER splicing into start of rootChunk.
     if isFennel then
-        local subopts = { nval = 1, target = target }
         local subscope = makeScope(rootScope.parent)
         if rootOptions.requireAsInclude then
             subscope.specials.require = requireSpecial
         end
-        local targetForm = list(sym('.'), sym('package.preload'), mod)
-        -- splice "or" statement in so it uses existing package.preload[modname]
-        -- if it's been set by something else, allowing for overrides
-        local forms = list(sym('or'), targetForm, list(sym('fn'), sequence()))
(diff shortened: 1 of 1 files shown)


technomancy:
@bakpakin I'm not sure I follow the subtlety of what's going on with `subChunk` vs `tempChunk` in the original implementation; do you think you could give this a once-over before merging?

jaawerth:
Due to the wackiness of `include`'s possible regressions and edge cases, I'd suggest we add tests in
- `test/misc.lua` testing nested `include`s (or nested `require` with `--require-as-include`), even if it's just comparing the compiled Lua output with the expected Lua output, making sure this nesting doesn't occur. We could even play around with the `require`/`include` order to test for the deduping issue that was fixed yesterday
- a test in `test/cli.fnl` that checks `fennel --version` when there's another version on `package.path`.

technomancy:
Wow, yeah geez. Looking at this more, there's definitely no way we should do anything more without some tests in place. There's just so many ways this can go wrong.

jaawerth:
After discussing with @technomancy, I've force-pushed a new implementation to this branch to keep it all in one PR. Nesting looks good now, and it's now indenting inlined Lua modules correctly as well as Fennel ones. Now we just need:
- A review of the new implementation by someone other than me :grin:
- A bunch of new tests to cover the various possible `include` regressions and edge cases.

bakpakin:
I’m quite wary of any implementation that is going to try and reformat Lua source code at all. For example, this will probably incorrectly mess with Lua multi-line strings. The first previous implementation was the way it was for a reason - without a Lua parser, the only good way to make sure we don’t change things is to emit the Lua exactly.

I think best solution here is to simply find a way to not emit metadata if that was the problem.

jaawerth:
@bakpakin, sorry if I'm missing something, but can you point to the changes in the diff you think are problematic?

> I’m quite wary of any implementation that is going to try and reformat Lua source code at all. For example, this will probably incorrectly mess with Lua multi-line strings. The first previous implementation was the way it was for a reason - without a Lua parser, the only good way to make sure we don’t change things is to emit the Lua exactly.
> 
> I think best solution here is to simply find a way to not emit metadata if that was the problem.

Avoiding the emission of metadata is (or at least was supposed to be) what this new version does; I believe the only change to the Lua-source-module side of things should be that, instead of manually string-interpolating the Lua module body into our generated function,

```lua
-- old way
emit(tempChunk, target .. ' = ' .. target .. ' or function()\n' .. s .. '\nend', ast)	
```

we're instead emitting
- the opening `package.preloaod["mymodule"] = package.preload["mymodule"] or function()`
- then the empty `subChunk`
- then `end`

to `tempChunk`, and if the module isn't Fennel code, emitting the Lua body directly into `subChunk`. That mostly just let me consolidate the control logic a bit, and let the chunks part of Lua emission handle the indentation the way it normally would.

The one formatting change I did make was using `:gsub` to strip trailing newlines when reading the source file's text, because I noticed it was embedding an unnecessary extra newline due to how `io.read` includes a file's terminating `\n` in the returned text. This *should* just mean going from
```lua
package.loaded["mymodule"] = package.loaded["mymodule"] or function()
  -- lua code here

end
```
to
```lua
package.loaded["mymodule"] = package.loaded["mymodule"] or function()
  -- lua code here
end
```
That said, we can ditch the `:gsub` if that's the bit you think might be problematic.

bakpakin:
Ah, I misread that pattern. Otherwise, I think the general idea should be sound. The previous logic is from (I believe) when we didn’t use package.preload and before functions had metadata. I agree on adding testing though.

jaawerth:
Sounds good! For future clarity, I'll also mention the reason for that `:gsub` pattern in a comment since it isn't obvious why it's there without comparing the output with and without it.

technomancy:
Awesome; let's get this merged and we can follow it up with some tests.

PR #287, opened 2020-05-22
Encode sequences in a way that's compatible with storing source data.

### LINKED ISSUE #286: [0.4.0] AST not applying SEQUENCE_MT to seq literals
It looks like sequential table literals are no longer being given `SEQUENCE_MT` in the ast, which is breaking the `sequence?` function in the compiler env, giving it false positives. This could cause issues when validating input to macros that accept e.g. a binding vector.

### Expected behavior (correct in 0.3.2)
```clojure
(macro seq-literal? [expr] (sequence? expr))
(seq-literal? [:a :b :c])
; => [:a :b :c]
```
### Actual (in 0.4.0 and current master)
```clojure
(macro seq-literal? [expr] (sequence? expr))
(seq-literal? [:a :b :c])
; => false
```

technomancy:
Using SEQUENCE_MT as the same metatable for every single sequence ever
makes it so that you can't store file/line data on them. Instead we
set the metatable to a new table that includes SEQUENCE_MARKER.

Also adds the sequence constructor function to compiler scope.

Fixes #286

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -108,7 +108,7 @@ local EXPR_MT = { 'EXPR', __tostring = deref }
 local VARARG = setmetatable({ '...' },
     { 'VARARG', __tostring = deref, __fennelview = deref })
 local LIST_MT = { 'LIST', __tostring = listToString, __fennelview = listToString }
-local SEQUENCE_MT = { 'SEQUENCE' }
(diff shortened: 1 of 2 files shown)


technomancy:
Thanks; I think you're right that a weak table is a better fit for this; we can switch it up in the next release.

PR #289, opened 2020-05-23
Use subscopes in for/each.

### LINKED ISSUE #288: for and each leak scope
```lisp
(for [y 0 2] nil)
(each [x (pairs [])] nil)

(match abc
  [x y] (print x y))
```

In the `match` section here, `x` and `y` should be fresh variables, because the locals in the loops above should not "escape" to the top-level scope. However, that's not actually the case in the expansion; the pattern variables are unified with the top-level `x` and `y`:

```lisp
(let [(_0_) abc]
  (if (and (= (type _0_) "table") (= (. _0_ 1) x) (= (. _0_ 2) y))
      (let [] (print x y))))
```

The correct expansion (which you get if you remove the `for` and `each`) binds `x` and `y` to fresh values:

```lisp
(let [(_0_) abc]
  (if (and (= (type _0_) "table") (not= nil (. _0_ 1)) (not= nil (. _0_ 2)))
      (let [x (. _0_ 1) y (. _0_ 2)]
        (print x y))))
```

technomancy:
Fixes #288.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -1904,28 +1904,29 @@ SPECIALS["each"] = function(ast, scope, parent)
     local iter = table.remove(binding, #binding) -- last item is iterator call
     local destructures = {}
     local newManglings = {}
+    local subScope = makeScope(scope)
     local function destructureBinding(v)
         if isSym(v) then
-            return declareLocal(v, {}, scope, ast, newManglings)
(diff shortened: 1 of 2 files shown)


jaawerth:
Nice catch!

PR #291, opened 2020-05-25
Searcher package config

### LINKED ISSUE #290: Invalid escape sequence when windows paths are in fennel.path when using macros
When paths in `fennel.path` table contain backslashes (`\`), they propagate through `searchModule` function. These paths appear in the source file if `doQuote` function is called. The function only wraps the path with quotes, without any additional processing. Backslash in literal string behaves as starting of an escape sequence, which leads to  `invalid escape sequence near ''C:'` error (`filename='C:\abc'`).

Workaround: before compiling change all backward slashes to forward slashes in `fennel.path`.
Possible fix: use `format("%q")` or any other alternative in `doQuote` to sanitize paths.

jaawerth:
Fixes #290

For cross-platform compat, `package.config` contains info about what to use as a dir separator, a delimiter in the package.path list, what to use as a placeholder, etc. This change uses the first three `package.config` values in `searchModule`, which should allow "\" in `fennel.path` to work on Windows.

## Recalculating every time?

I initially put this logic in `searchModule` so that it's recalculated every time. However, because these values are set internally by the Lua VM during its "compile time" (see [Lua reference](https://www.lua.org/manual/5.3/manual.html#pdf-package.config) and [the Lua C source that sets it](https://github.com/lua/lua/blob/master/loadlib.c#L744-L746)), changes to `package.config` during runtime don't appear to impact `require` or `package.searchpath`; local testing in Lua seems to verify this.

## Testing

We can't really add automated tests for this unless we make circleci rerun tests in a Windows VM, but before merging we should see if someone can test this branch on a Windows machine to ensure it actually does address #290. 

**EDIT:** Even without windows, we can probably at least test the fix to `doQuote` that I just pushed by adding a dir with a backslash to `fennel.path`. I'm out of time to work on this until tonight or tomorrow, but if someone wants to add such a test and get this merged, feel free.

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -2637,15 +2637,20 @@ module.repl = function(options)
     return eval(replsource, { correlate = true }, module, internals)(options)
 end
 
-local function searchModule(modulename, pathstring)
-    -- use package.config to process package.path (e.g. for windows compat)
+-- have searchModule use package.config to process package.path (windows compat)
+local pkgConfig
+do
     local cfg = string.gmatch(package.config, "([^\n]+)")
     local dirsep, pathsep, pathmark = cfg() or '/', cfg() or ';', cfg() or '?'
-    local escapedSep = escapepat(pathsep)
-    local pathsplit = string.format("([^%s]*)%s", escapedSep, escapedSep)
-    modulename = modulename:gsub("%.", dirsep)
-    for path in string.gmatch((pathstring or module.path)..escapedSep, pathsplit) do
-        local filename = path:gsub(escapepat(pathmark), modulename)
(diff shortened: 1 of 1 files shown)


jaawerth:
Whoops, `make ci` is failing for this in circleci, but wasn't for me locally, I believe because of some `FENNEL_PATH` and `LUA_PATH` values I had set on my shell. I probably won't have time to revisit until this evening and correct whatever I got wrong, but I imagine it's something simple that I missed.

technomancy:
Just tested this with a macro in a directory that contained a backslash, and it failed on master but worked on this branch, so it looks like it does the job. Nice work.

PR #295, opened 2020-06-12
implement `with-open` macro

jaawerth:
This introduces a new macro based on clojure's [with-open](https://clojuredocs.org/clojure.core/with-open). Its semantics are the same as `let`, only it expects every binding to be a file handle (or at least something with a `:close` method), and it will automatically invoke `:close` on every binding either after the body has finished or upon encountering an error.

The motivation for this was mainly to reduce the boilerplate of invoking `(file:close)`, with the added benefit that even if there's an error, the file descriptors will *still* be closed immediately instead of whenever the vm gets around to garbage-collecting the file handle's binding.

~~I went with `let-open` for the name because the semantics are so close to that of `let`, which should keep cognitive debt down when it comes to what the semantics are.~~

**Update**: It's now disallowing destructuring, so I've renamed it to `with-open` to get away from implying that the bindings semantics are just like `let`.

### Example

```clojure
(print (with-open [fin (io.open :somefile)]
         (fin:read :*a)))
```
expands to
```clojure
(print
  (let [fin (io.open "somefile")]
    (fn close-handlers_0_ [ok_0_ ...]
      (: fin "close")
      (if ok_0_ ... (error ... 0)))
    (close-handlers_0_ (xpcall (fn [] (fin:read "*a"))
                               (. (or package.loaded.fennel debug) "traceback")))))
```

### DIFF
--- a/fennel.lua
+++ b/fennel.lua
@@ -2954,6 +2954,18 @@ Same as ->> except will short-circuit with nil when it encounters a nil value."
          (assert body1 "expected body")
          `(if ,condition
               (do ,body1 ,...)))
+ :let-open (fn [closable-bindings ...]
+              "Like `let`, but invokes (v:close) on every binding after evaluating the body.
+The body is evaluated inside `xpcall` so that bound values will be closed upon
+encountering an error before propagating it."
+              (let [bodyfn    `(fn [] ,...)
+                    closer    `(fn close-handlers# [ok# ...] (if ok# ... (error ... 0)))
+                    traceback `(. (or package.loaded.fennel debug) :traceback)]
+                (for [i 1 (# closable-bindings) 2]
+                  (table.insert closer 4 `(: ,(. closable-bindings i) :close)))
+                `(let ,closable-bindings
+                   ,closer
+                   (close-handlers# (xpcall ,bodyfn ,traceback)))))
  :partial (fn [f ...]
             "Returns a function with all arguments partially applied to f."
             (let [body (list f ...)]
--- a/fennel.lua
+++ b/fennel.lua
@@ -2954,7 +2954,7 @@ Same as ->> except will short-circuit with nil when it encounters a nil value."
          (assert body1 "expected body")
          `(if ,condition
               (do ,body1 ,...)))
- :let-open (fn [closable-bindings ...]
+ :with-open (fn [closable-bindings ...]
               "Like `let`, but invokes (v:close) on every binding after evaluating the body.
 The body is evaluated inside `xpcall` so that bound values will be closed upon
 encountering an error before propagating it."
@@ -2962,9 +2962,10 @@ encountering an error before propagating it."
                     closer    `(fn close-handlers# [ok# ...] (if ok# ... (error ... 0)))
                     traceback `(. (or package.loaded.fennel debug) :traceback)]
                 (for [i 1 (# closable-bindings) 2]
+                  (assert (sym? (. closable-bindings i))
+                    "with-open only allows symbols in bindings")
                   (table.insert closer 4 `(: ,(. closable-bindings i) :close)))
-                `(let ,closable-bindings
-                   ,closer
+                `(let ,closable-bindings ,closer
                    (close-handlers# (xpcall ,bodyfn ,traceback)))))
  :partial (fn [f ...]
             "Returns a function with all arguments partially applied to f."
--- a/reference.md
+++ b/reference.md
@@ -244,6 +244,39 @@ Example:
   (table.concat c ",")) ; => "3,4,5,6"
 ```
 
+### `let-open` bind and auto-close file handles
+
+*(Since ??? TODO: add version before release)*
+
+While Lua will automatically close an open file handle when it's garbage collected,
+GC may not run right away; `let-open` ensures handles are closed immediately, error
+or no, without boilerplate.
+
+The usage is the same as `let`, only every binding should be a file handle or other value
+with a `:close` method. After executing the body, or upon encountering an error, `let-open`
+will invoke `(value:close)` on every bound variable before returning the results.
+
+The body is implicitly wrapped in a function and run with `xpcall` so that all bound
+handles are closed before it re-raises the error.
+
+Example:
+
+```fennel
+; Basic usage
+(let-open [fout (io.open :output.txt :w) fin (io.open :input.txt)]
+  (fout:write "Here is some text!\n")
+  (fin:line)) ; => first line of input.txt
+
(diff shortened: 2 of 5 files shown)


fennel.lua
```diff
@@ -2946,6 +2946,18 @@ Same as ->> except will short-circuit with nil when it encounters a nil value."
          (assert body1 "expected body")
          `(if ,condition
               (do ,body1 ,...)))
+ :let-open (fn [closable-bindings ...]
+              "Like `let`, but invokes (v:close) on every binding after evaluating the body.
+The body is evaluated inside `xpcall` so that bound values will be closed upon
+encountering an error before propagating it."
+              (let [bodyfn    `(fn [] ,...)
+                    closer    `(fn close-handlers# [ok# ...] (if ok# ... (error ... 0)))
```
technomancy, line 2954: The use of `...` here is a little funny; doesn't the fact that `...` is used in the middle of the `if` and `error` calls mean that it will collapse down to a single value? Seems like we should use `val#` here instead maybe, or use `(values ...)` in the first branch of the `if`.

test/core.lua
```diff
@@ -344,6 +344,13 @@ local function test_macros()
         ["(-?>> :w (. {:w :x}) (. {:x :missing}) (. {:y :z}))"]=nil,
         ["(-?> [:a :b] (table.concat \" \"))"]="a b",
         ["(-?>> \" \" (table.concat [:a :b]))"]="a b",
+        -- let-open should close all handles and return result of body
+        ["(var (fh1 fh2) nil) [(let-open [f1 (io.tmpfile) f2 (io.tmpfile)]\
+          (set [fh1 fh2] [f1 f2]) (f1:write :asdf) (f1:seek :set 0) (f1:read :*a))\
+          (io.type fh1) (io.type fh2)]"]={"asdf", "closed file", "closed file"},
+        -- let-open should propogate errors and still close the file handles
+        ["(var fh nil) (local (ok msg) (pcall #(let-open [f (io.tmpfile)] (set fh f)\
+          (error :bork!)))) [(io.type fh) ok (msg:match :bork!)]"]={"closed file", false, "bork!"},
```
technomancy, line 353: We should add a test for multiple values here.

reference.md
```diff
@@ -244,6 +244,39 @@ Example:
...
+with a `:close` method. After executing the body, or upon encountering an error, `let-open`
+will invoke `(value:close)` on every bound variable before returning the results.
+
+The body is implicitly wrapped in a function and run with `xpcall` so that all bound
+handles are closed before it re-raises the error.
+
+Example:
+
+```fennel
+; Basic usage
```
technomancy, line 265: Top-level comments should have two semicolons rather than one.

reference.md
```diff
@@ -244,6 +244,39 @@ Example:
...
+  (fout:write "Here is some text!\n")
+  (fin:line)) ; => first line of input.txt
+
+; This demonstrates that the file will also be closed upon error.
+(var fh nil)
+(local (ok err)
+  (pcall #(let-open [file (io.open :test.txt :w)]
+            (set fh file) ; you would normally never do this
+            (error :whoops!))))
+(io.type fh) ; => "closed file"
```
technomancy, line 276: Do you think maybe we could show trying to write to the file here and show that it has an error instead?

fennel.lua
jaawerth: So, this could either use a comment explainer or a more verbose change for better clarity, but the fact that it collapses down to a single value is actually why it behaves correctly. I originally had `(let [msg ...] (error msg 0))`, but `(error ... 0)` is semantically equivalent without the extra variable binding, so I typed it that way.

Definitely happy to change it, I just figured in the name of keeping emitted code as overhead-free as possible that this implementation makes the most sense; definitely comment-worthy, though.

reference.md
jaawerth: You mean assume the write errored instead of causing an error in the code?

test/core.lua
jaawerth: That it returns multiple values? Or.. hmm, actually this raises a valid point. What `let-open` doesn't account for here is destructuring or multiple-value bindings like

```clojure
(let-open [(f1 f2) (values (io.open :a) (io.open :b))] xyz)
```

I can either add handling for that to make it continue to work, or I could make a no-destructuring assertion. Thoughts?

technomancy: Oh good point. I was thinking of multiple values in the body, but I hadn't considered multiple values in the binding. I think it's OK skip destructuring as long as we call it out as unsupported in the docs, and maybe add an explicit error message asserting that each binding is a `sym?`.

fennel.lua
technomancy: Oh! I wasn't thinking of the `error` part; I was assuming that the `(if ok# ...` part would collapse down to a single value. But since it's in a tail position I guess it does actually work? I assumed it wouldn't since it's not at the end of the form, but now that I think about it more, it should be the tail position that's more relevant here. Just tested it in the repl and this appears to be the case, so never mind!

reference.md
technomancy: Yeah, show that attempting to write to `file` results in an error outside the `let-open` form maybe? Just an idea since it might be clearer than the `io.type` function; that might be less familiar to folks, but it's fine as-is too.

test/core.lua
jaawerth: So I initially agreed with this, but after letting it simmer for a week, I don't feel great about the `let-open` name if we're going to make the binding semantics stray from those of `let`. Clojure's `with-open` also only allows syms in the name position, no destructuring, so doing that here would track, but I think that flips my position back to pro `with-open` over `let-open` for the fennel equivalent. What do you think?

reference.md
jaawerth: Hrm, well the `io.type` is there mainly to demonstrate that even when there *are* errors, the file-handlers are still closed; demonstrating that the errors still propagate was more of a bonus.

I'm happy to change it that makes these points clearer, but my gut says we should still demonstrate both behaviors. Any ideas that would demonstrate both with more clarity? Off the top of my head, instead of explicitly raising an error I could have something like `(file:nonexisted-method "this will raise an error")`.

We could always merge it as-is and clarify the reference later, too.

test/core.lua
jaawerth: Went ahead and added a `sym?` assertion and renamed it to `with-open`!

reference.md
technomancy: Yeah, we can change this after merging. Everything else looks great. Nice work.

PR #297, opened 2020-06-22
Pseudo modules

technomancy:
This somewhat-invasive change breaks out the `fennel.lua` file into four distinct "pseudo-modules":

* utils
* parser
* compiler
* specials

This is done by creating top-level functions that return tables but isolate scope, ensuring that the dependencies between the various sections of the codebase are clear. But apart from the clarity advantages, this also opens the door to moving towards making the compiler self-hosting; we can write Fennel code which loads up each of the individual modules without porting them all to Fennel at the same time. First the outer bit gets ported (using `include` to splice the existing Lua implementations directly into the compiler output) and eventually on a one-by-one basis we can rewrite each module in Fennel, instead of requiring it to be done in one fell swoop.

This patch changes a lot of code, but it doesn't change the structure of the codebase or add any functionality; it should be an entirely pure refactor.

Note: it's highly recommended to read this diff with [whitespace disabled](https://github.com/bakpakin/Fennel/pull/297/files?w=1) because so much indentation changes.

Edit: I'm not really sure what kind of review I'm expecting here because the diff is just so enormous;  I guess the main things it's helpful to get eyes on are:

* do these four sections make sense as a way to break down the code?
* do the "exports" for each section make sense?
    * [utils](https://github.com/bakpakin/Fennel/blob/pseudo-modules/fennel.lua#L281)
    * [parser](https://github.com/bakpakin/Fennel/blob/pseudo-modules/fennel.lua#L617)
    * [compiler](https://github.com/bakpakin/Fennel/blob/pseudo-modules/fennel.lua#L1613)
    * [specials](https://github.com/bakpakin/Fennel/blob/pseudo-modules/fennel.lua#L2690)
* there are a few things that [just didn't fit cleanly into any pseudo-module](https://github.com/bakpakin/Fennel/pull/297/commits/da8a70f)


jaawerth:
Sorry for the delay on reviewing this, but it looks great!

PR #298, opened 2020-06-26
implement varargs $... in hashfn

### LINKED ISSUE #209: varargs in hashfn
There's currently no way to refer to `...` in hashfn's. Initially this made sense to me, as they're really just meant as shorthand sugar for very simple functions, but there are a few instances where it would really come in handy, as well as some powerful usecases I think we're missing out on.

### Function composition
While we could introduce a `comp` macro, we wouldn't really need it if we had this. Assuming, for the case of argument, hashfn varargs take the form of `$&`, left-right composition is as simple as
```clojure
#(-> $& func1 func2 func3)
```
There's still an issue with multiple returns making this harder, but if we introduced `as->` (separate issue forthcoming) or another means to concisely limit and manipulate function args, that too cleans up nicely.

### Wrapping other functions
One of the primary usecases for hashfn is to wrap other functions or operators, and it obviates the need for some kind of placeholder special for doing positional partial application. This is somewhat limited by the lack of `...`.

## What should it be?

- `$...` would make the most sense but is somewhat awkward.
- `$&` would be in line with clojure's use of `%&`, but it isn't quite as obvious since fennel sticks with `...` instead of using `&` in normal functions.
- `$$` is easy to remember.
- Just use `...`. Works because `...` isn't lexically scoped, but may make the syntax more confusing to someone picking up the language?
- other?

jaawerth:
Resolves #209 

## varargs

First and foremost, this implements `$...` for varargs in hashfns.

To prevent possibly-confusing behaviors mentioned in [this comment in #209](https://github.com/bakpakin/Fennel/issues/209#issuecomment-623048237), for now `$...` is mutually exclusive with positional `$`,`$1`,`$2`, etc.
In the future, we can easily relax this constraint and replace it with an assertion that, for example, `$...` can only be used if positional arguments are contiguous.

## walkTree

This also introduces a new internal function, `walkTree`. Not only can this potentially be used elsewhere in the compiler, but we may also be able to expose it on the compiler env for macro purposes. Hashfns are now using it to recursively walk the AST, transforming all occurrences of `$...` into `...`.

```lua
walkTree(ast[2], function(node, idx, parentNode)
    if isSym(node) and deref(node) == '$...' then
        parentNode[idx], fScope.vararg = VARARG, true
    end
    return isList(node) or isTable(node) -- return is used to decide whether to branch
end)
```


technomancy:
Nice work; this looks great!

PR #300, opened 2020-06-27
Two new options for fennelview: :empty-as-square and :table-edges.

technomancy:
The first one just makes empty tables emit as [] instead of {}.

The second one makes it so k/v tables emit as

```
{:a 1
 :b 2
 :c 3}
```

rather than

```
{
 :a 1
 :b 2
 :c 3
}
```

Mostly just opening this as a PR to get suggestions for possible better names for these options.

Both these will make `fnlfmt` a lot more pleasant to use.


jaawerth:
👍

PR #305, opened 2020-07-10
add Lua 5.4 tests to CI

jaawerth:
I noticed that Lua 5.4 had hit Debian Testing, so I tried changing our build job to use the `debian:testing` docker image and added lua5.4 to the tests. Works!

Did this as a PR mainly to give @technomancy veto power over making the change from `debian:buster-slim` to `debian:testing`.


technomancy:
Nice work; thanks!

PR #306, opened 2020-07-12
Self-host the compiler

technomancy:
Each former pseudo-module (utils, parser, compiler, and specials) gets moved to a real module, but still written in Lua.

Edit: this now includes a complete port for full self-hosting. The only remaining Lua is the old pre-selfhosted bootstrap files (under `old/`), the `test/init.lua` wrapper that bootstraps, and the generated compiler output.

The `src/fennel.fnl` file was hand-ported, the other `src/fennel/*.fnl` files were compiled with antifennel, along with some manual changes to add in comments, convert local->let, and a few other minor changes.

The repl gets moved from an inline string into a proper file.

The old implementation is now in old/fennel.lua and is used to bootstrap the new one. The fennel.lua file is now a generated file.


technomancy:
This still needs more changes; it passes tests but the output of the standalone `fennel` script has a bug that isn't present in `fennel.lua` where macros don't get loaded.

old/fennel.lua
```diff
@@ -2693,7 +2693,7 @@ local specials = (function()
             -- Compile the forms into subChunk; compiler.compile1 is necessary for all nested
             -- includes to be emitted in the same root chunk in the top-level module
             for i = 1, #forms do
-                local subopts = i == #forms and {nval=1, tail=true} or {}
+                local subopts = i == #forms and {nval=1, tail=true} or {nval=0}
```
technomancy, line 2696: This one is actually a legitimate bug in the old compiler that we didn't find until we tried to use `include` on Fennel itself.

Makefile
```diff
@@ -17,28 +20,26 @@ testall: fennel
...
-# For the time being, avoid chicken/egg situation thru the old Lua launcher.
-LAUNCHER=$(LUA) old_launcher.lua
+# Avoid chicken/egg situation using the old Lua launcher.
+LAUNCHER=$(LUA) old/launcher.lua --add-fennel-path src/?.fnl --globals "_G,_ENV"
 
 # Precompile fennel libraries
-%.lua: %.fnl fennel.lua
-	 $(LAUNCHER) --globals "" --compile $< > $@
+fennelview.lua: fennelview.fnl fennel.lua ; $(LAUNCHER) --compile $< > $@
+fennelfriend.lua: fennelfriend.fnl fennel.lua ; $(LAUNCHER) --compile $< > $@
```
technomancy, line 32: Should we continue to keep these generated files in the repo, or should we move them to the downloads section of the web site? (We can address this later on.)

jaawerth: Agree we should address it later on, but I think we should move to an all-or-nothing approach - either we leave all the generated files out, or check them all in and mark them as generated in `.gitattributes`.

If we do the latter, we should also add a CI check to ensure that at the very least all the timestamps match, either that or do a checksum of the built files in the repo against some fresh-built equivalents.

Makefile
```diff
@@ -17,28 +20,26 @@ testall: fennel
...
 
 # All-in-one pure-lua script:
-fennel: launcher.fnl fennel.lua fennelview.lua fennelfriend.lua fennelbinary.fnl
+fennel: launcher.fnl $(SRC) fennelview.lua fennelfriend.lua fennelbinary.fnl
 	echo "#!/usr/bin/env $(LUA)" > $@
-	$(LAUNCHER) --globals "" --require-as-include --metadata --compile $< >> $@
+	$(LAUNCHER) --no-metadata --require-as-include --compile $< >> $@
```
jaawerth, line 37: Why the change to `--no-metadata`? Was this to get around the problem with circular requires?

My only issue there is that it's nice to be able to use `doc` on fennelview from within the standalone launcher.

src/fennel.fnl
```diff
@@ -0,0 +1,110 @@
...
+(fn dofile* [filename options ...]
+  (let [opts (utils.copy options)
+        f (assert (io.open filename :rb))
+        source (f:read :*all)]
+    (f:close)
+    (set opts.filename filename)
+    (eval source opts ...)))
+
+;; The public API module we export:
+(local mod {
```
jaawerth, line 51: Unnecessary line break? Minor, I know!

src/fennel.fnl
```diff
@@ -0,0 +1,110 @@
...
+            :dofile dofile*
+            :version "0.5.0-dev"
+
+            :repl repl
+            })
+
+;; This is bad; we have a circular dependency between the specials section and
+;; the evaluation section due to require-macros/import-macros, etc. For now
+;; stash it in the utils table, but we should untangle it
+(set utils.fennelModule mod)
```
jaawerth, line 90: What if we make some things lazy-loadable by setting a `__newindex` metamethod on modules that lazy-loads the circular things? That way it doesn't try to invoke the code with the circular dep until the other dep has already loaded.

That said, I think we could merge this and play with that in a subsequent PR.

src/fennel.fnl
```diff
@@ -0,0 +1,110 @@
...
+
+;; Load the built-in macros from macros.fnl.
+(let [builtin-macros (eval-compiler
+                       (with-open [f (assert (io.open "src/fennel/macros.fnl"))]
+                         (.. "[===[" (f:read "*all") "]===]")))
```
jaawerth, line 95: This question may become moot when I do the incremental version anyway, and I'm going to play with a couple angles to it after we merge, but: Instead of doing this, and the below `eval`, did you try just using `specials["require-macros"]` here? That _may_ resolve the windows string issue as well.

Just curious, but I'm going to tackle this part post-merge anyway.

old/fennel.lua
jaawerth: Nicely spotted!

Makefile
technomancy: Since a CI run will remake the generated files, maybe we can add a final step to the CI which fails the build if the generated files have changed according to git.

src/fennel.fnl
technomancy: I think the problem is that `require-macros` would load it in during compile-time for the compiler rather than compile-time for the program being compiled. But maybe there's a way to make it work anyway?

src/fennel.fnl
technomancy: I'm not sure I follow but yeah, if you want to take a shot at it after the merge that'd be great. I don't really like the way this looks right now.

Makefile
jaawerth: True, we could just add something like `git diff --quiet`, which will exit with a `1` status if there are changed files. Or there's probably a circleci orb for it, but I'm not super familiar with that ecosystem.

technomancy: Cool; added in aaa9f8d.

PR #311, opened 2020-07-22
Add Homebrew install to setup.md

SeekingMeaning:
https://formulae.brew.sh/formula/fennel


technomancy:
Looks good! Maybe we can expand this into a section on "using your OS's package manager" but this is a great start. Thanks.

Issue #310, opened 2020-07-19
Get rid of camelCased identifiers in the compiler


technomancy:
When we ported the compiler to being self-hosted, we kept all the identifier names the same as the old Lua implementation, (using camelCase) but what we really want is to use kebab-case everywhere we can.

This applies to everything in the compiler *except* the public API, (the `mod` table in the `src/fennel.fnl` file) which we can't change for compatibility reasons.

This can be done one file at a time for the most part. If anyone wants to get started hacking on the compiler, this could be a good way to dip your toe in the water since it's very straightforward but it will touch a lot of the code in simple ways.

Note that changing the name of locals can be done in a way that just affects a single file, but when changing the module exports at the bottom of the file, the consumers of that module will need to be updated as well.

technomancy:
Fixed in 0c109e6.

PR #312, opened 2020-07-26
[RFC] convert to kebab-case (#310)

atweiden:
Backwards compatibility aliases provided.


technomancy:
Thanks! However, it looks like this change isn't quite ready--it fails to compile. The following patch fixes the compilation error: https://p.hagelb.org/0001-Avoid-eval-compiler-collision.patch.html

However even after that's fixed there are failures in the test suite that we will need to resolve before this can be applied:

```
$ make test
lua test/init.lua
1..33
# Started on Sat Jul 25 19:54:07 2020
ok     1	test-macros
ok     2	test-comparisons
ok     3	test-parsing
ok     4	test-functions
ok     5	test-calculations
ok     6	test-method_calls
ok     7	test-booleans
ok     8	test-destructuring
ok     9	test-conditionals
ok     10	test-edge
ok     11	test-match
ok     12	test-loops
ok     13	test-if
ok     14	test-core
ok     15	test-fennelview
ok     16	test-hashfn
ok     17	test-mangling
ok     18	test-quoted-source
ok     19	test-quote
ok     20	test-include
ok     21	test-env-iteration
ok     22	test-traceback
ok     23	test-runtime-quote
ok     24	test-leak
ok     25	test-global-mangling
ok     26	test-empty-values
not ok 27	test-no-undocumented
#   fennel.lua:437: can't look up doc with metadata disabled.
not ok 28	test-docstrings
#   fennel.lua:437: can't look up doc with metadata disabled.
ok     29	test-fennelview
ok     30	test-suggestions
not ok 31	test-failures
#   test/failures.fnl:7: Expected compiling (let [next #(next $)] print) to fail.
#   expected: false, actual: true
>> 
>> 
not ok 32	test-completion
#   test/repl.fnl:34: cannot resume dead coroutine
ok     33	test-cli
# Ran 33 tests in 0.334 seconds, 29 successes, 1 failure, 3 errors
Makefile:12: recipe for target 'test' failed
make: *** [test] Error 1
```

technomancy:
Oh, looking thru the patch again, I think I might know what's causing the failures. The changes to the fields in the `opts` or `options` tables will need to be reverted; those will cause backwards-incompatibility issues. I think the best thing to do is to keep those as camelCase for the purposes of this patch. After this is merged we can add make further changes to accept those in kebab-case or camelCase if we decide we want to go that way.

Thanks for taking the time to put this together!

atweiden:
Tests are passing.

I had intended to squash the commits.

technomancy:
Nice; the latest changes have tests passing for all versions of Lua except 5.1. It looks like some changes to the traceback code is causing compatibility issues:

```
Testing lua 5.1:
...........F........E............
Tests with errors:
------------------
1) test-traceback
fennel.lua:1901: bad argument #1 to 'format' (string expected, got nil)
stack traceback:
	fennel.lua:1901: in function 'traceback_frame'
	fennel.lua:1930: in function 'traceback'
	test/mod/tracer.fnl:3: in function 'inner'
	test/mod/tracer.fnl:7: in function 'tracer'
	test/misc.fnl:5: in function 'test-traceback'

Failed tests:
-------------
1) test-macros
test/core.fnl:84: (var fh nil) (local (ok msg) (pcall #(with-open [f (io.tmpfile)] (set fh f)
          (error :bork!)))) [(io.type fh) ok (msg:match :bork!)]
expected: {"closed file", false, "bork!"}
actual: {"closed file", false}
stack traceback:
	test/core.fnl:84: in function 'test-macros'

Ran 33 tests in 0.376 seconds, 31 successes, 1 failure, 1 error
```

This is probably caused by the fact that `(string.format "%s" nil)` is an error in Lua 5.1 but not in newer versions. I can take a look later to try and fix it but if you get to it first that'd be cool too. =)

technomancy:
Thanks for this! I tracked down the problem to the use of `short-src`; it turns out that we must use `short_src` because that's what Lua uses: https://www.lua.org/pil/23.1.html

As far as I know this is the only place in all of Lua that uses snake_case! I had no idea. Anyway I fixed that in 3b665c2 and merged it on it. Thanks for your contribution.

atweiden:
You're welcome. Fennel is quite an exceptional project, and I'd like to contribute further.

Granted, I had intended to squash the commits -- primarily to keep the todo out of git history -- but that was mostly for vanity I suppose.

> After this is merged we can add make further changes to accept those in kebab-case or camelCase if we decide we want to go that way

Thoughts on kebab-casing opts?

technomancy:
> Thoughts on kebab-casing opts?

I'd like to see this as well (with backwards-compatibility); I just didn't want to do it in the same commit since that one was already massive and touching so many files that I wanted to get it merged as quickly as possible to reduce opportunity for conflict.

For the `opts` table, perhaps we should have a function that "canonicalizes" every entry in the opts table by converting camelCase opts into kebab-case opts and signaling an error if both versions are provided. We would have to ensure that this function was called to canonicalize the opts for every possible entry point where an opts table could be passed in. That probably means that the canonicalizer function needs to be in the `fennel.utils` module.

Does that make sense?

atweiden:
Cool. I'll check it out. However, I perhaps enjoyed a meal with one too many fennel seeds added last night: and accidentally lacerated my left index finger on a food processor blade. Timing is slightly dependent on my ability to type with two index fingers, but yeah, camel case has got to go. But I'm also looking at wiring up [Olical/aniseed](https://github.com/Olical/aniseed) macros for `defn` and friends. It's [so beautiful](https://github.com/atweiden/vim-fennel#screenshot).

Issue #314, opened 2020-07-31
tutorial: what characters can be used when naming things?


uvtc:
Somewhere near the beginning of the tutorial it should mention what characters can be used when naming things (such as functions, variables, etc.). That is, it's a surprise later when you see things can be named with a question mark in them. Can you use exclamation marks too? (Hm, why would you?) How about a trailing single-quote-mark (prime)? Any reason to use an underscore (maybe for let bindings that you aren't concerned with ... maybe this can be mentioned later when destructuring is explained).

Also maybe an indented note about naming conventions in Fennel would be helpful?

technomancy:
Great point! This would be a good addition to the tutorial; we should add it.

uvtc:
Ok, will do, but could you first please tell me:

  * can you use `!` in a name, and is there a reason to? (I think some langs may use this to indicate mutation, but I don't think Lua nor Fennel does...)
  * can you use single quote as a prime, and should it even be suggested? (as opposed to just coming up with another name, rather than `old-name'`)
  * is an underscore typically used when you don't care about the value, as in `(let [[foo _ bar] some-arr] ...)`?
  * is an underscore used in some other way? For example, some langs use them to indicate private functions, or functions that are not expected to be called outside of a given module.
  * It looks like you can *start* an identifier with a question mark (for example, in the section on function arity and lambda). That looks unusual to me. What is the convention for prepending a `?` vs appending one?

technomancy:
> Ok, will do, but could you first please tell me:

Sorry; didn't mean to say that you should be the one to do it. But if
you want to, I won't object. =)

>   * can you use `!` in a name, and is there a reason to? (I think some
>   langs may use this to indicate mutation, but I don't think Lua nor
>   Fennel does...)

It's a legal character, yes. It's often used to indicate side-effects,
but it's not a convention to always use it for anything side-effecting.

>   * can you use single quote as a prime, and should it even be
>   suggested? (as opposed to just coming up with another name, rather
>   than `old-name'`)

This is currently not supported; quote and backquote are treated as
form-level quoting even if it's at the middle or end of a symbol.

I think we could look into changing this, but that's how it currently works.

>   * is an underscore typically used when you don't care about the
>   value, as in `(let [[foo _ bar] some-arr] ...)`?

Yep!

>   * is an underscore used in some other way? For example, some langs
>   use them to indicate private functions, or functions that are not
>   expected to be called outside of a given module.

No, it is just the unused binding.

[...]

PR #315, opened 2020-07-31
tutorial: note on naming identifiers

uvtc:
For #314 .

More complete details could go into the reference later on.

### DIFF
--- a/tutorial.md
+++ b/tutorial.md
@@ -27,7 +27,7 @@ simply an anonymous value.
 > must start with a letter or underscore. You can also use the question mark
 > (typically for functions that return a true or false, ex., `at-max-velocity?`).
 > Underscores (`_`) are often used to name a variable that we don't plan
-> on using (seen further down when we cover "destructuring").
+> on using.
 
 The argument list is provided in square brackets. The final value is
 returned.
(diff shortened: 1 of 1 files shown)


technomancy:
Thanks; this looks good, except it refers to the destructuring section when mentioning underscore, but the destructuring section doesn't mention anything about underscores. Did you mean to add another section there?

uvtc:
No, I didn't mean to add another section for this. I see that destructuring is covered in the "back to tables just for a bit" section, but there's no destructing subsection, so I just mentioned it but didn't provide a link. I think there needs to be *something* after that comment about using underscores, otherwise the reader will think "variables you don't intend to use?? Bah, that makes no sense", but I don't want to throw out fancy terminology without at least a little tidbit to let the reader know, "don't worry, we'll get to that in a bit".

technomancy:
I guess I'm still not following; I don't see what the connection is between underscores and destructuring. Underscores are used anywhere that you can bind a value, regardless of whether it's being destructured or not. For example, there is no destructuring here:

```
(fn [a _ msg]
  (when a
    (print msg)))
```

Nor is there here:

```
(let [f (io.open filename)
      _ (assert f)
      contents (f:read :*all)]
  (consume contents))
```

uvtc:
Ah! Whoops. I see what you mean. Ok, removed that little bit.

technomancy:
Cool; thanks!

Issue #309, opened 2020-07-18
setup guide: move repl/readline section & fix ToC


jaawerth:
Just filing this so I/we don't forget to make a couple tweaks to the setup guide.

I noticed that in the [setup guide](setup.md),
- The section about adding/configuring readline support in the REPL is under the "Adding Fennel support to your text editor" section. We should move it to somewhere more appropriate
- The top-level ToC link for the readline section is broken

technomancy:
I've added a pandoc-generated TOC on all guides here: https://git.sr.ht/~technomancy/fennel-lang.org/commit/4ae7c6576518cde97bc172de128a9de3147e7704

but the existing markdown file has TOC entries in it that are now redundant.

technomancy:
The latest changes to the setup guide move it so readline setup comes right after "Adding Fennel support to your text editor" instead of under it. I think that covers this issue, right?

PR #318, opened 2020-08-01
setup.md: tidied up sections

uvtc:
Removed redundant embedded mini-ToC's.
Fixed heading levels so they fall under the correct section.
Heading levels now are "one lower" so they render right in the ToC at the top.
Also some minor markdown formatting tweaks (ex. list indents)

### DIFF
--- a/setup.md
+++ b/setup.md
@@ -7,30 +7,17 @@ edit configuration files in a UNIX-like environment.
 **Note**: Fennel can be used in non-UNIX environments, but those environments
 will not be covered in this document.
 
-This document consists of the following sections:
 
-* [Downloading Fennel](#downloading-fennel)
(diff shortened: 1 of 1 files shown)


uvtc:
Fixes #309

uvtc:
Regarding indenting lists and sublists, I follow the rule that list content (regardless of marker) starts at 4-spaces in. And sublists 4 *more* spaces in, and so on. This:

  * lets you nest lists arbitrarily deep (at least on Pandoc),
  * makes everything look uniform (including having extra paragraphs in one list item), and
  * ensures you don't accidentally end up with code blocks where you thought list content would be.

technomancy:
Looks good; thanks.

PR #324, opened 2020-08-18
add --lua flag to launcher

jaawerth:
This introduces a new flag that re-executes the given CLI command with the provided options in a child process using the supplied Lua executable via `os.execute`. It allows CLI users to run the launcher in an arbitrary lua VM without resorting to such shellisms as `luaX.Y $(which fennel)`, decoupling the launcher from its host runtime.

For example,
```shell
fennel --lua luajit -e '_VERSION'
```
will invoke the equivalent of
```shell
luajit path/to/fennel -e '_VERSION'`
```
## Flag name

I'd be open to being more explicit with `--lua-version`, but I kind of like the brevity of `--lua`. I don't feel strongly either way, so I figured I'd mention it before we make it legacy.

## Tests

The tests for this flag only run under `make testall`, since that's the only time we can safely assume the host environment has every lua version available.

Initially, I was letting the test suite for every lua version run the `--lua` test for every *other* version, giving us a total of 20 `io.popen` invocations, which in turn invoke `os.execute`, but this led to a rather significant slowdown.

Instead, I have each test for a given Lua executable simply test the flag for the next Lua executable in the list. It does add some complex logic, and there's still some slowdown (about half a second on my machine), but that's better than quadrupling the delay.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,7 @@ build: fennel
 test: fennel.lua fennel
 	$(LUA) test/init.lua
 
+testall: export FNL_TESTALL = 1
 testall: export FNL_TEST_OUTPUT ?= text
 testall: fennel
 	@printf 'Testing lua 5.1:\n'  ; lua5.1 test/init.lua
--- a/changelog.md
+++ b/changelog.md
@@ -4,6 +4,7 @@
 
 * Fix a bug where method calls would early-evaluate their receiver.
 * Fix a bug where multi-arity comparisons would early-evaluate their arguments.
+* Add `--lua` CLI flag for specifying a custom Lua command/executable. (#324)
 
 ## 0.5.0 / 2020-08-08
 
--- a/src/launcher.fnl
+++ b/src/launcher.fnl
@@ -21,6 +21,7 @@ Run fennel, a lisp programming language for the Lua runtime.
   --no-metadata           : Disable function metadata, even in REPL
   --correlate             : Make Lua output line numbers match Fennel input
   --load FILE (-l)        : Load the specified FILE before executing the command
+  --lua LUA_EXE           : Run in a child process with LUA_EXE (experimental)
   --no-fennelrc           : Skip loading ~/.fennelrc when launching repl
   --compile-binary FILE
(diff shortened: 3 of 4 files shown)


technomancy:
> I'd be open to being more explicit with `--lua-version`, but I kind of
> like the brevity of `--lua`

I prefer --lua over --lua-version because the latter makes me think that
it would take something like "5.3" or "5.1" rather than the name of a command.

>   --lua LUA_EXE           : Run in a child process with LUA_EXE (experimental)

Can we call it LUA_CMD instead of LUA_EXE?

> (for [i (# arg) 1 -1]
>   (match (. arg i)
>     "--lua" (handle-lua i)
>     "--no-searcher" (do (set options.no_searcher true)
>                         (table.remove arg i))

One problem I can see with the current handling of the arguments is that
we check for --lua inside the loop where we walk thru all of `arg`. But
the problem is that this loop also makes destructive changes to
`arg`. So if --lua is the first thing it checks, you're fine, but if
it's the last arg, then several other flags may have been removed and
placed into the options table.

So I think in order for this to be done consistently, we should check
for --lua separately first before we do the regular arg loop.

Other than that it looks great!

jaawerth:
@technomancy
> I prefer --lua over --lua-version because the latter makes me think that it would take something like "5.3" or "5.1" rather than the name of a command.

Great point! We're specifying an executable path, not a version.

> Can we call it LUA_CMD instead of LUA_EXE?

For sure, that definitely sounds better.

> One problem I can see with the current handling of the arguments is that we check for --lua inside the loop where we walk thru all of `arg`. But the problem is that this loop also makes destructive changes to `arg`. So if --lua is the first thing it checks, you're fine, but if it's the last arg, then several other flags may have been removed and placed into the options table. So I think in order for this to be done consistently, we should check for --lua separately first before we do the regular arg loop. Other than that it looks great!

Good call. The flag parsing fragility occurred to me earlier when I first whipped this up but I forgot to explore its breaking points.

Your suggestion of handling `--lua` first is a good fix for the purposes of this PR, but longterm I'd like to move args-handling into a separate module anyway, and make it a bit more robust across the board by e.g. parsing it into a table and making it easier to concisely define validations for missing/incompatible args.

Issue #325, opened 2020-08-19
fenneview doesn't put the table id before its body when showing the first instance of a table that is present multiple times.


xoich:
If for example we give this table to the viewer

```lua
t = {}
t.t = t
```
`inspect.lua` will show this:

```
<1>{
  t = <table 1>
}
```
while fennel only this:
```
{
  :t #<table 1>
}
```
---
 Another example:

```lua
a = {}
t = {a,a}
```

`inspect.lua` (insecting table `t`):
```
{ <1>{}, <table 1> }
```
fennel:
```
[{} #<table 2>]
```

technomancy:
This is fixed in 944342e.

PR #327, opened 2020-08-19
 Fix issue #325 with Fennelview

xoich:


technomancy:
>          (when (and (< 1 (or (. self.appearances t) 0)) self.detect-cycles?)
>            (puts self "<" id ">"))

Thanks!

This looks pretty good, except for the bit above it uses the inspect.lua
style of printing tables rather than the style already used elsewhere by
fennelview; let's change it to be consistent with the #<table 1> style
seen higher up in the function.

technomancy:
Thanks for this. I merged this and went ahead and made the change we talked about in chat to use `@` instead in 289d55f.

PR #337, opened 2020-12-16
Fennel support in Vis, Textadept, and Howl editors

lepotic:


technomancy:
Thanks! Could you provide links to the editors in question? I've never heard of them, and I expect many others won't know what they are.

technomancy:
Cool; thanks! It's great to hear about Fennel showing up in more places I've never even heard of. =D

PR #346, opened 2021-04-12
Test that pattern matches before destructuring

mrichards42:
This patch fixes a bug where the `where` clause in `match` binds values (including destructuring table patterns) before testing that the pattern actually matches.

Without this patch, the included test errors.

A minimal example:

```clj
>> (match nil [x y z] :table _ :not-a-table)
"not-a-table"

>> (match nil (where [x y z] (not= nil x)) :table _ :not-a-table)
runtime error: attempt to index local '_0_0' (a nil value)
```


technomancy:
Great find again; thanks.

PR #347, opened 2021-04-12
Fix match multi-value destructuring with guards

mrichards42:
I noticed while debugging #346 that `match` was always binding at least 3 values when I used a `where` clause, because there are usually 3 elements in the `(where ...)` list. This patch changes the multi-value calculation so that it checks the actual values being destructured, even when nested in a `where` clause.

```clj
;; before
>> (macrodebug (match (values 1 2 3 4 5) (where (a b c d e) (= 1 a)) a))
(let [(_0_ _1_ _2_) (values 1 2 3 4)] ; <- missing _3_ and _4_
  (if ... match clauses ...))

;; after
>> (macrodebug (match (values 1 2 3 4 5) (where (a b c d e) (= 1 a)) a))
(let [(_0_ _1_ _2_ _3_ _4_) (values 1 2 3 4 5)]
  (if ... match clauses ...))

### DIFF
--- a/src/fennel/macros.fnl
+++ b/src/fennel/macros.fnl
@@ -364,10 +364,13 @@ introduce for the duration of the body if it does match."
   "How many multi-valued clauses are there? return a list of that many gensyms."
   (let [syms (list (gensym))]
     (for [i 1 (length clauses) 2]
-      (if (list? (. clauses i))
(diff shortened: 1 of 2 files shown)


technomancy:
Sorry I missed this earlier. This is a great catch; thanks for submitting the fix.

PR #350, opened 2021-05-06
Lua Primer typo fixes and LFS link

glinesbdev:
Fix a typo (missing word) in the `table.sort` section as well as adding a reference link to Lua File System

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -1,10 +1,5 @@
 # Summary of user-visible changes
 
-## 0.9.4 / 2021-05-05
-
-* Fixed a typo in the Lua Primer page
-* Added a reference link to the Lua File System
-
 ## 0.9.3 / ???
 
 * Add `fennel.syntax` function describing built-ins.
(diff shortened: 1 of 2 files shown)


changelog.md
```diff
@@ -1,5 +1,10 @@
 # Summary of user-visible changes
 
+## 0.9.4 / 2021-05-05
+
+* Fixed a typo in the Lua Primer page
+* Added a reference link to the Lua File System
+
```
jaawerth, line 7: For the changelog, we typically only include user-facing changes to Fennel itself (the compiler, the cli, etc), since the changelog mainly exists for people to keep track of differences between releases; no need for changelog updates in the docs, since the changelog is, in a way, a part of the docs 😉

jaawerth (changes requested):
Thanks for the revisions! Looks good, I'd just ask that you remove the changelog update and we're good to merge.

changelog.md
glinesbdev: Sounds good. Will keep in mind for the future. Thanks!

jaawerth (approved):

PR #352, opened 2021-05-10
Fix destructuring kv tables inside values

mrichards42:
(I'm very much not familiar with the compiler, but I at least tracked the error down to this line, and tests pass, so I figure there's a decent chance this is the right fix)

When symmeta is added in destructure-values, we have to check that the
value is in fact a sym (not a table). Any names declared in the table
destructure have their symmeta added in the recursive call to
destructure1.

This manifests as a compiler error:

    (let [({:x 10}) {:x 10}] x) => Error: table index is nil

... from setting a table key to (utils.deref {:x 10}), which is nil.

Though note that sequential table destructuring accidentally worked,
since utils.deref is just (. sym 1), or in this case, `x`:

    (let [([x]) [10]] x) => 10

This appears to be a regression from b43fe48.

### DIFF
--- a/src/fennel/compiler.fnl
+++ b/src/fennel/compiler.fnl
@@ -722,7 +722,8 @@ which we have to do if we don't know."
         (compile-top-target left-names)
         (when declaration
           (each [_ sym (ipairs left)]
-            (tset scope.symmeta (utils.deref sym) {:var isvar})))
+            (when (utils.sym? sym)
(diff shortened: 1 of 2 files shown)


technomancy:
Good catch; thanks! This is one of the hairiest parts of the compiler that I haven't gotten around to refactoring and breaking down into more bite-sized pieces yet, but your fix seems pretty clearly correct.

PR #360, opened 2021-06-04
Fixes sentence

jgarte:


sogaiu:
The original text looks ok to me.

jgarte:
> The original text looks ok to me.

I think for the original to be gramatically correct then one other fix would be to use the present participle of interact:

> ... property of the table itself but determined by a given piece of code *interacting* with the table ...

A change is required either way for it to be grammatically correct.

sogaiu:
Oh, sorry, I failed to notice "interacts with the table." as part of the original sentence.

I agree the original suggested change makes more sense.

technomancy:
Applied; thanks!

PR #369, opened 2021-07-03
Add accumulation macro and its tests

### LINKED ISSUE #367: Reduce (or fold) macro
Fennel provides nice macros `collect` and `icollect`, which enable us to map iterator to a table succinctly.
It might be too much for Fennel's minimalism I guess, but could be nice to have another macro to reduce/fold iterators.
Something like this:
```
(macro reduce1 [iter-tbl accum-expr]
  (let [acc (sym :acc)]
    `(do (var ,acc ,(table.remove iter-tbl 1))
         (each ,iter-tbl
           (match ,accum-expr
             v# (set ,acc v#)))
         ,acc)))
; (reduce1 [1 _ n (ipairs [2 3 4 5])]
;   (* acc n)) ;=> 120

; or

(macro reduce2 [iter-tbl accum-expr]
  (let [acc (table.remove iter-tbl 1)]
    `(do (var ,acc ,(table.remove iter-tbl 1))
         (each ,iter-tbl
           (match ,accum-expr
             v# (set ,acc v#)))
         ,acc)))
; (reduce2 [state "" k v (pairs {:a 1 :b 2 :c 3 :d 4})]
;   (when (= (% v 2) 0)
;     (.. state k))) ;=> "bd"
```

mnacamura:
Close #367.

Following @technomancy's advice (thanks!), I fixed the macro to not skip `nil` in each accumulation step.

Because `reduce` and `fold` are frequently used in third-party libraries, I named it `accumulate`.
I'm still unsure about the name though. `accumulate` might be too long to type so its shorthand `accum` or `reduce` or `fold` could be better. Anyway suggestions are welcome.

### DIFF
--- a/changelog.md
+++ b/changelog.md
@@ -12,6 +12,7 @@ Backwards-incompatible changes are **marked in bold**.
 * Make macro tables shadow runtime tables more consistently
 * Add `,complete foo` repl command
 * Add `fennel.syntax` function describing built-ins
+* Add `accumulate` macro
 
 ## 0.9.2 / 2021-05-02
 
--- a/reference.md
+++ b/reference.md
@@ -812,6 +812,28 @@ value into a table is a no-op.
 Like `each` and `for`, the table comprehensions support an `:until`
 clause for early termination.
 
+### `accumulate` iterator accumulation
+
+Run through an iterator and performs accumulation, similar to `fold`
+and `reduce` commonly used in functional programming languages.
+Like `collect` and `icollect`, it takes an iterator binding table
+and an expression as its arguments. The difference is that in
+`accumulate`, the first two items in the binding table are used as
+an "accumulator" variable and its initial value.
+For each iteration step, it evaluates the given expression and
+the returned value becomes the next accumulator variable.
+`accumulate` returns the final value of the accumulator variable.
+
+Example:
+
+```fennel
+(accumulate [avg 0
+             i n (ipairs [1 2 3 4])]
+  (let [/i (/ i)]
+    (+ (* avg (- 1 /i)) (* n /i))))
+;; -> 2.5
+```
+
 ### `values` multi-valued return
 
 Returns multiple values from a function. Usually used to signal
--- a/src/fennel/macros.fnl
+++ b/src/fennel/macros.fnl
@@ -149,7 +149,7 @@ returns
        (tset tbl# (+ (length tbl#) 1) ,value-expr))
      tbl#))
 
-(fn accum* [iter-tbl accum-expr ...]
+(fn accumulate* [iter-tbl accum-expr ...]
   "Accumulation macro.
 Similar to `collect` and `icollect`, it takes a binding table and an
 expression as its arguments.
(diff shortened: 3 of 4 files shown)


andreyorst:
`accumulate` is a bit long indeed, but I think it's OK. However, this right here:

>Because reduce and fold are frequently used in third-party libraries, I named it accumulate.

Indicates that we're stepping into dangerous territory, where we might really want to have our own replaceable standard library to avoid future problems with making it harder to use third-party libraries, which may include names identical to the names of macros supplied with Fennel, especially given our overshadowing semantics that will freak out because of that. This is an unrelated issue, but I think this is a good indication that it should be addressed.

I've tested your macro and it seem to work correct. I would suggest to add a bit more tests, with empty tables for example, to ensure that body is not called and initial value is returned as is. Maybe tests with `:until` aslo make sense.

mnacamura:
I added more tests, @andreyorst thanks!

In addition, I shortened the macro's name to `accum`; shorter looks nicer for me after all..
<image src="https://user-images.githubusercontent.com/45770/124370954-b775ab00-dcb7-11eb-8da6-a936993f6a4d.png" width=320px>

andreyorst:
I think `accumulate` was better, because there is lesser chance of full name like that being used in third party library, but I think the final word is behind @technomancy :)

technomancy:
> Because reduce and fold are frequently used in third-party libraries

The other reason is that when you see `reduce` you might assume that it's a function rather than a macro. Since this works differently we should name it differently even if it's used to solve the same problem.

Yeah, I'm leaning more towards `accumulate` myself. We have very few abbreviations in Fennel; almost all the ones we have are inherited from elsewhere. (`fn` and `var`) The only Fennel-specific abbreviation I can think of is `tset` and I don't really like it much. =)

Of course, the advantage of `accum` is that it doesn't push code over to the right as much so that less indentation would be needed for the iterator section. One other alternate name that is shorter and doesn't require abbreviation is `inject` which is what reduce is called in Smalltalk. I'm somewhat on the fence; what do others think about that name?

andreyorst:
> One other alternate name that is shorter and doesn't require abbreviation is `inject` which is what reduce is called in Smalltalk. I'm somewhat on the fence; what do others think about that name?

I'm unfamiliar with this name in other languages. I feel that `accumulate` conveys the intention more clear

mnacamura:
My thought on `inject`:

#### Pros
* Shorter than `accumulate`, 60% of length.
* Smalltalk also has `collect` which already Fennel has so it feels coherent.
#### Cons
* In Smalltalk, I think its syntax much helps clarifying its intention (`obj inject: initialValue into: lambdaFunction`). In our case, we don't have such at least in the current form.

andreyorst:
I think `inject` makes sense in the context of Smalltalk because of OOP, where you really inject into an object, which we don't do in Fennel, as we just `set`.

technomancy:
Yeah, thinking about it more I agree `inject` is not the right name here. Let's stick with `accumulate`.

technomancy:
Thanks! I've applied the patch for this.

PR #370, opened 2021-07-05
added unless macro

eko234:
Maybe fennel can use a unless macro, what do you think?


technomancy:
Thanks, but Fennel is a fairly minimalistic language so I feel like there needs to be more usefulness to a builtin macro than just letting you omit a single `not` form.

PR #371, opened 2021-07-06
Evaluate args to partial only once.

mrichards42:
Simple literal values (strings, numbers, booleans, and nil) can be
inlined into the function expansion. Everything else (e.g. tables,
symbols, function calls) is bound once in a let form.

Discussed in IRC:

```clj
(macrodebug (partial + (math.random 10)))
;; before
(fn [...] (+ (math.random 10) ...))
;; after
(let [_1_ (math.random 10)] (fn [...] (+ _1_ ...)))


;; though simple values are still inlined
(macrodebug (partial + 10))
;; before
(fn [...] (+ 10 ...))
;; after
(let [] (fn [...] (+ 10 ...))
```

### DIFF
--- a/src/fennel/macros.fnl
+++ b/src/fennel/macros.fnl
@@ -152,10 +152,23 @@ returns
 (fn partial* [f ...]
   "Returns a function with all arguments partially applied to f."
   (assert f "expected a function to partially apply")
-  (let [body (list f ...)]
-    (table.insert body _VARARG)
-    `(fn [,_VARARG]
-       ,body)))
+  (let [bindings []
+        args []]
+    (each [_ arg (ipairs [...])]
+      (if (or (= :number (type arg))
+              (= :string (type arg))
+              (= :boolean (type arg))
+              (= `nil arg))
+        (table.insert args arg)
+        (let [name (gensym)]
+          (table.insert bindings name)
+          (table.insert bindings arg)
+          (table.insert args name))))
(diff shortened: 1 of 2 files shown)


src/fennel/macros.fnl
```diff
@@ -152,10 +152,23 @@ returns
...
   "Returns a function with all arguments partially applied to f."
   (assert f "expected a function to partially apply")
-  (let [body (list f ...)]
-    (table.insert body _VARARG)
-    `(fn [,_VARARG]
-       ,body)))
+  (let [bindings []
+        args []]
+    (each [_ arg (ipairs [...])]
+      (if (or (= :number (type arg))
```
technomancy, line 158: I think this can actually be shortened to `(not= :table (type arg))` since that covers symbols as well as other tables; does that sound right to you?

mrichards42: I think it would have to be `(or (not= :table (type arg)) (= 'nil arg))` since it looks like nil is parsed as a symbol, but I'm good with that overall. I assume it isn't possible for the parser to return userdata? That's the only other exception I can think of.

technomancy: Oh true; I forgot that nil is still a symbol at this point.

Userdata and coroutines are the other types but yeah, they are not valid inside an AST.

technomancy:
Went ahead and applied as-is; thanks!

PR #373, opened 2021-07-08
Add :preprocess option to fennel.view

### LINKED ISSUE #356: Provide a way to specify conversion fuctions for custom types for fennel.view
E.g. `awful.key` from Awesome WM Lua API creates an object of a custom type where `type(k)` is `"key"`, so using `(print (fennel.view k))` prints something like:
```
#<key: 0x563c7303a538>
```
At the same time, a key instance has properties that I know about and that I can access. It would be nice to have a way to provide a custom conversion to `fennel.view` from a custom type into one of the Lua types to allow pretty printing of instances of such types.

p-himik:
It allows for pretty-printing of any userdata with custom type.
E.g. in Awesome WM a key object has `type(obj) = "key"`.

Fixes #356

### DIFF
--- a/src/fennel/view.fnl
+++ b/src/fennel/view.fnl
@@ -255,6 +255,7 @@ as numeric escapes rather than letter-based escapes, which is ugly."
           ;; main serialization loop, entry point is defined below
           (let [indent (or indent 0)
                 options (or options (make-options x))
+                x (if options.preprocess (options.preprocess x options) x)
                 tv (type x)]
             (if (or (= tv :table)
                     (and (= tv :userdata)
@@ -290,6 +291,9 @@ Can take an options table with these keys:
   lengths
 * :max-sparse-gap (integer, default 10) maximum gap to fill in with nils in
   sparse sequential tables.
+* :preprocess (function) if present, called on x (and recursively on each value
+  in x), and the result is used for pretty printing; takes the same arguments as
+  `fennel.view`
 
 The `__fennelview` metamethod should take the table being serialized as its
 first argument, a function as its second argument, options table as third
@@ -375,5 +379,10 @@ results regardless of nesting:
 
 Note that even though we've only indented inner elements of our table
 with 10 spaces, the result is correctly indented in terms of outer
-table, and inner tables also remain indented correctly."
+table, and inner tables also remain indented correctly.
+
+When using the `:preprocess` option, avoid modifying any tables in-place in the
+passed function. Since Lua tables are mutable and provided to `:preprocess`
+without copying, any modification done in `:preprocess` will be visible outside
+of `fennel.view`."
   (pp x (make-options x options) 0))
--- a/test/core.fnl
+++ b/test/core.fnl
@@ -417,7 +417,14 @@
                               (fennel.view styles {:prefer-colon? true})
                               (fennel.view styles {:prefer-colon? false})]
                              {:one-line? true})"
-               "[\":colon \\\"quote\\\" \\\"depends\\\"\" \":colon \\\"quote\\\" :depends\" \":colon \\\"quote\\\" \\\"depends\\\"\"]"}]
+               "[\":colon \\\"quote\\\" \\\"depends\\\"\" \":colon \\\"quote\\\" :depends\" \":colon \\\"quote\\\" \\\"depends\\\"\"]"
+               ;; :preprocess
+               "((require :fennel.view) [1 2 3] {:preprocess (fn [x] x)})"
+               "[1 2 3]"
+               "((require :fennel.view) [1 2 3] {:preprocess (fn [x] (if (= (type x) :number) (+ x 1) x))})"
+               "[2 3 4]"
+               "((require :fennel.view) [[] [1] {:x [] [] [2]}] {:preprocess (fn [x] (if (and (= (type x) :table) (= (next x) nil)) :empty-table x))})"
+               "[\"empty-table\" [1] {:x \"empty-table\" :empty-table [2]}]"}]
     (each [code expected (pairs cases)]
       (l.assertEquals (fennel.eval code {:correlate true :compiler-env _G})
                       expected code))


andreyorst:
>
>
> It allows for pretty-printing of any userdata with custom type.
> E.g. in Awesome WM a key object has type(obj) = "key".
>
> Fixes #356 <https://github.com/bakpakin/Fennel/issues/356>
>
I think the documentation should be more clear on the fact, that this is
really necessary only when working with userdata. I understand that it can
be used with anything, but data transformations for ordinary tables is
already possible by implementing custom `__fennelview` metamethod.

Also please add this change to the changelog with a comprehensive
description.

Thanks for working on this! I wasn't aware of custom types in userdata when
I was working on fennelview rewrite.

>

src/fennel/view.fnl
```diff
@@ -255,6 +255,7 @@ as numeric escapes rather than letter-based escapes, which is ugly."
           ;; main serialization loop, entry point is defined below
           (let [indent (or indent 0)
                 options (or options (make-options x))
+                x (if options.preprocess (options.preprocess x options) x)
```
andreyorst, line 258: Have you considered for `prerocess` not to be a function, but a `preprocessors` table?

``` clojure
{:key (fn [ud] ...)
 :other-type (fn [ud] ...)}
```

This way way, we will be able to specify a conversion function tied to a particular type. Then, `pp` function instead of converting everything using the same function would instead call `type` on the `x` and search if the resulting type exists in `options.preprocessors`:

``` clojure
(let [indent (or indent 0)
      options (or options (make-options x))
      x (match (. options.preprocessors (type x))
          f (f x options)
          _ x)
      tv (type x)]
  (if (or (= tv :table)
...
```

I think this will allow a more fine control over userdata printing, and will help avoid confusion and errors when all data is preprocessed for no real reason.

p-himik: > I think this will allow a more fine control over userdata printing

How is it better than a single `if` within the function itself?

What if I want to trim `:password` field from _all_ tables, regardless how nested they are?

> will help avoid confusion and errors when all data is preprocessed for no real reason

I think a better way to do that would be to give an example of the `preprocess` function with a clear default return value with a comment describing why it's done that way.

p-himik:
> I understand that it can be used with anything, but data transformations for ordinary tables is already possible by implementing custom `__fennelview` metamethod.

Have you read this comment in the original discussion? https://github.com/bakpakin/Fennel/issues/356#issuecomment-843279448

andreyorst:
> I understand that it can be used with anything, but data transformations
> for ordinary tables is already possible by implementing custom
> __fennelview metamethod.
>
> Have you read this comment in the original discussion? #356 (comment)
> <https://github.com/bakpakin/Fennel/issues/356#issuecomment-843279448>
>

That's why I explicitly said "for ordinary tables".

>

andreyorst:
>
> How is it better than a single if within the function itself?
>
This provides an interface. And it makes function far less complected - the
type matching is not function's job, the data transformation is.

> What if I want to trim :password field from *all* tables, regardless how
> nested they are?
>
{:key (fn [ud] ...)
 :table (fn [t] (hide-password ...)}

This will transform any tables regardless of how nested they are, and as a
bonus you will not need to modify an existing function for awesome wm key
userdata transformer.

> will help avoid confusion and errors when all data is preprocessed for no
> real reason
>
> I think a better way to do that would be to give an example of the
> preprocess function with a clear default return value with a comment
> describing why it's done that way.
>
The clear default for preprocess function is the value it was given. This
is the same as not applying any function at all.

5 simple functions working with their 5 respecting data types is less
complect than 1 function working with 5 data types.

>

src/fennel/view.fnl
andreyorst: >
> How is it better than a single if within the function itself?
>
This provides an interface. And it makes function far less complected - the
type matching is not function's job, the data transformation is.

> What if I want to trim :password field from *all* tables, regardless how
> nested they are?

``` clojure
{:key (fn [ud] ...)
 :table (fn [t] (hide-password ...)}
```

This will transform any tables regardless of how nested they are, and as a
bonus you will not need to modify an existing function for awesome wm key
userdata transformer.

> will help avoid confusion and errors when all data is preprocessed for no
> real reason
>
> I think a better way to do that would be to give an example of the
> preprocess function with a clear default return value with a comment
> describing why it's done that way.
>
The clear default for preprocess function is the value it was given. This
is the same as not applying any function at all.

5 simple functions working with their 5 respecting data types is less
complect than 1 function working with 5 data types.

>

p-himik:
> That's why I explicitly said "for ordinary tables".

That comment describes why `__fennelview` is impractical for even ordinary tables. There are data structures I know nothing about, and I can't imagine someone wanting to walk the entire tree to just add that field when `preprocess` can do just that without any hassle.
`__fennelview` is for something known and controlled. `preprocess` is for the rest.

> the type matching is not function's job, the data transformation is

The function's job is preprocessing of any kind, of any data. What it ends up doing - that's the user's concern. As a user, I don't want to be limited by something intended for development purposes. One can imagine a case where I _don't_ want to match on any type - I want the function to run for _all_ types. For example, when that function comes from a library someone else has written - imagine that Awesome WM has its own type conversion function that deals with all its C types. Or what if I just want to log all data types - exactly because I don't know them?

> {:key (fn [ud] ...)
 :table (fn [t] (hide-password ...)}

Ah, of course. Still waking up, sorry about that.

> 5 simple functions working with their 5 respecting data types is less
complect than 1 function working with 5 data types.

I would agree if it was something that ends up as a part of a complex system, where such a difference could matter.
But here we're talking about a dev-time facility. It should be easy and useful, it shouldn't get in the developer's way with its pedantic interfaces. And if for some obscure reason, someone decides to use `fennel.view` in their release code as a part of the intended app's usage - well, they can wrap it in something that does that table dispatch.

andreyorst:
> That comment describes why `__fennelview` is impractical for even ordinary tables. There are data structures I know nothing about, and I can't imagine someone wanting to walk the entire tree to just add that field when preprocess can do just that without any hassle.
> `__fennelview` is for something known and controlled. preprocess is for the rest.

Yes, I agree that `__fennelview` is impractical to implement for something like hiding a password field, that's defenitively a `preprocess` job. What I've meant is that if you're creating some kind of data structure like [`ordered-set`]() it's your job to provide a sensible data transformation via `__fennelview` so it could be printed without modifying `options` table in the `view` method, e.g. in the REPL.

> The function's job is preprocessing of any kind, of any data. What it ends up doing - that's the user's concern. As a user, I don't want to be limited by something intended for development purposes. One can imagine a case where I _don't_ want to match on any type - I want the function to run for _all_ types. For example, when that function comes from a library someone else has written - imagine that Awesome WM has its own type conversion function that deals with all its C types. Or what if I just want to log all data types - exactly because I don't know them?

This can be implemented as a `:default` type that will run for any type unless it has a specific function, which is defined separately. This is similar to how multimethods work in Clojure. But I agree that this actually makes it more complicated to use.

Why I think that this matters. The `pp` function is called on each and every element of the data we're pretty printing. These elements can both come from iterating over a table inside the `view` function, or artificially, when `__fennelview` metamethod calls `pp` function on the elements which can be anything. I think this is not reasonable to preprocess everything with the same function, as this can possible break on user-defined types made with sparse tables for example. Yes this can still break if you specify the `:table` type in the `preprocessors` table, but it's unlikely to happen that often.

> I would agree if it was something that ends up as a part of a complex system, where such a difference could matter.
> But here we're talking about a dev-time facility. It should be easy and useful, it shouldn't get in the developer's way with its pedantic interfaces. And if for some obscure reason, someone decides to use `fennel.view` in their release code as a part of the intended app's usage - well, they can wrap it in something that does that table dispatch.

Of course, I've originaly asked "have you considered" question, so I'm not imposing this solution or forcing it in any way, it's just a discussion :) My main concern is that calling it on everything may be problematic, but it's the problem of the one who calls it, yes. If we're ok with this, I think we should explicitly mention it somewhere. You can add a more elaborated section in the bottom of the docstring, explaining how to write a more or less generic function for data transformation, mentioning that it should return the original data if no transformation is needed.

p-himik:
I see, sounds good!

> I think this is not reasonable to preprocess everything with the same function, as this can possible break on user-defined types made with sparse tables for example.

Can you elaborate?
Or even better - create a small example of how one can break things so that I could directly include it in the docstring as an example of what not to do, what to be aware of?

andreyorst:
> Can you elaborate?
> Or even better - create a small example of how one can break things so that I could directly include it in the docstring as an example of what not to do, what to be aware of?

Well, because tables are mutable, you should avoid `preprocess` modifying things in place -- it should to create a copy of the table first. Otherwise this is a huge problem:

``` clojure
(local passwd-data {:password "foobar"})
(view passwd-data) ;; prints  `{:password "foobar"}`
(view passwd-data {:preprocess (fn [t] hide-the-pass)}) ;; prints `{:password "******"}`
(view passwd-data) ;; prints  `{:password "******"}`
(use-password passwd-data) ;; wrong password...
```

And proxy tables can be problematic here for example, because if you create a copy you need to create a copy of its metatable as well. Otherwise it can print worng because `__fennelview` metamethod is lost. Deep copying is hard, and can be problematic as we recursivly call `pp` on a deep table, which does recursive deep copies of various fields for modification.

I'm not sure I have any exact examples in mind. I'll try to think of some in the evening, but the rough idea is above.

src/fennel/view.fnl
jaawerth: I think adding this option is a good idea as an alternative when `__fennelview` either can't be used or would be unwieldy. @p-himik do you mind adding a couple test cases? It's just a one-line change, but we'll want some coverage for it just to prevent any future regressions and the like.

### preprocess fn vs preprocessors table
I would prefer the the `options.preprocess` as written over cooking in the type dispatching:

- It keeps the preprocessors composable - if you want to combine preprocessing logic you can just compose functions together instead of composing multiple functions in multiple tables
- The type dispatching in the function doesn't really add that much complexity if you need it (it's a very common, and clean, usecase of `match`), while not locking you into any specific patterns (you could also just do duck-typing).
- A preprocessor that appllies to multiple types and/or needs to do some sort of duck-typing or other conditional logic would take at least as much extra boilerplate with a type-dispatching preprocessors table as keeping the option simple and leaving type dispatch to the function itself.

src/fennel/view.fnl
```diff
@@ -290,6 +291,8 @@ Can take an options table with these keys:
   lengths
 * :max-sparse-gap (integer, default 10) maximum gap to fill in with nils in
   sparse sequential tables.
+* :preprocess (function) if present, it's called on each value in x and the
+  result is then used instead of x for pretty printing; takes the same arguments
```
jaawerth, line 295: I think this wording is a bit confusing as-is, since the proprocessor will be called only on X, and then later (potentially) on each value in X as fennelview recursively outputs the results. I think we'll want to explicitly mention the recursion in there, maybe something like:
```suggestion
* :preprocess (function) if present, called on x (and recursively on each value in x),
  and the result is used for pretty printing; takes the same arguments as fennel.view
```

jaawerth (changes requested):
Looks good to me! My only ask other than throwing in a couple test cases (which I mentioned in a comment thread but figured I'd also mention here to avoid derailing the other convo), I just think the docstring could use some minor tweaking for clarity (though we can always play with the wording later too).

technomancy:
I agree; the code looks good, but we need test cases as well as some more clarity in the docstring. Thanks.

technomancy:
I'd love to get this into a release this upcoming week if we could get the docstring and tests improved.

technomancy:
Thanks! The new docstring looks good. I've merged your patch.

PR #374, opened 2021-07-21
Fix lua special and add test case

sioonhho:
The lua special function doesn't produce what the website, or comments say it should.

This statement:
```
(each [_ v (ipairs (lua nil "{1,2,3}"))]
  (print v))
```

should print the contents of the table, but results in the following lua:
```
nil
for _, v in ipairs({1,2,3}) do
  print(v)
end
return nil
```

and the following lua runtime error: `unexpected symbol near 'nil'`.

Since the lua special doesn't compile its arguments, comparing to `nil` always results in false because it's comparing to the result from the parser, not the nil literal the compiler would output. Checking if the argument is a symbol, dereferencing it if it is, and comparing it to `:nil` instead of `nil` works correctly.


technomancy:
Nice catch; thanks! I've applied your patch.

PR #380, opened 2021-08-08
Reference.md tweaks

rktjmp:
- Small additions of links or cleanup.
- Changes `accumulate` example to something simpler.
- Adds very brief explanation of macros application.
- Adds return types to compiler environment functions.

Macro explanation uses `plus-10`, which might infer you should be replacing `plus-10` functions with macros. Not sure if that needs a note or not.


reference.md
```diff
@@ -1005,6 +1002,28 @@ macros into smaller scopes as well. Note that macros are a
 compile-time construct; they do not exist at runtime. As such macros
 cannot be exported at the bottom of a module like functions and other values.
 
+Briefly, a macro can be thought to "paste" an espression in its place. A macro
+should return a "quoted" expression (via `` ` ``), to be "pasted". That
```
andreyorst, line 1006: maybe instead of "paste" the "splice" term can be used? I think I saw "spliced in" in some macro docs for other lisps

rktjmp: Yeah I didn't super love "paste". Splice is a bit of an unusual term too, technically it's joining or inserting, not replacing.

Maybe just "replace", it's probably most explicit?

> When called, a macro replaces it's own expression with another expression, derived from macro.

Saying "expression" "expression" "derived" isn't so approachable, depending on whether "expression" is a term assumed to be "known" by the reader, but it's probably most clear that `x becomes y`?

>  Briefly, a macro can be thought to splice an expression in its place.

> When called (invoked?), a macro splices an expression in its place.

Also it should be *ex*pression not ~~espresso~~ *es*pression.

reference.md
```diff
@@ -1255,14 +1274,18 @@ can use `tostring` to get the name of a symbol.
 
 * `list` - return a list, which is a special kind of table used for code
 * `sym` - turn a string into a symbol
-* `list?` - is the argument a list?
-* `sym?` - is the argument a symbol?
-* `table?` - is the argument a non-list table?
+* `gensym` - generates a unique symbol for use in macros (also available via
+             `<var-name>#`).
```
andreyorst, line 1278: Gensym and autogensym are a bit different, so I think the additional info in the parentheses may be misleading, because `(gensym :foo)` produces `foo_N_`, where N is some unique number but `foo#` produces `foo_N_auto`, and the text may be understood as if the name generated with `gensym` can be available by using this name with `#` appended.

reference.md
```diff
@@ -1255,14 +1274,18 @@ can use `tostring` to get the name of a symbol.
...
+* `list?` - is the argument a list? Returns `argument` or `false`.
```
andreyorst, line 1279: Maybe this is a bit clearer?

```suggestion
* `list?` - is the argument a list? Returns its argument or `false`.
```

rktjmp: _the_ argument? That follows the previous sentence.

- `list?` is the argument a list? Returns the argument or `false`.

reference.md
rktjmp: Maybe separate them out?

* `gensym` - generates a unique symbol for use in macros, accepts an optional prefix string.
* `<symbol>#` - generates a unique symbol, derived from the given symbol (name?).

reference.md
andreyorst: > Maybe just "replace", it's probably most explicit?

Yeah, probably. My thinking was something like that:

> Briefly, a macro can be thought to splice an expression in its place.

Splice in sounds right, because there's `unquote-splicing` in other Lisps, which ATM is missing in Fennel.

But I guess somenting like "replacing the call of the macro with its returned quoted expression" seems the most clear from the perspective of someone who doesn't know macros in other Lisps.

reference.md
andreyorst: `<symbol>#` may be confused with REPL representation of some un-printable objects, like `#<function: 0x55937ccb5720>`

I guess it could be

* autogensym - a unique symbol generated when appending `#` to a symbol name inside quoted expression

But I'm open to other ideas

reference.md
```diff
@@ -1005,6 +1002,30 @@ macros into smaller scopes as well. Note that macros are a
 compile-time construct; they do not exist at runtime. As such macros
 cannot be exported at the bottom of a module like functions and other values.
 
+Broadly, a macro replaces its own expression with another.
+
+Macros should return a "quoted" expression (via `` ` ``) to be used as the
+replacement. Normally that expression would "unquote" a value (via `,`) but it
+does not have to. You may not "unquote" a value outside of a "quoted"
+expression.
```
rktjmp, line 1010: Probably simplest way to say it?

reference.md
```diff
@@ -1005,6 +1002,30 @@ macros into smaller scopes as well. Note that macros are a
...
+
+As a very simple example, the following macro will emit an expression that adds
+10 to a given argument.
+
+```fennel
+(macro plus-10 [arg]
+  `(+ 10 ,arg))
+; ^ quote the following expression, it will be emitted in place of (plus-10 ...)
```
rktjmp, line 1018: I think emit is a normal enough term

reference.md
```diff
@@ -1253,26 +1274,30 @@ from regular tables defined with square or curly brackets. Similarly symbols
...
 
-* `list` - return a list, which is a special kind of table used for code
-* `sym` - turn a string into a symbol
-* `list?` - is the argument a list?
-* `sym?` - is the argument a symbol?
-* `table?` - is the argument a non-list table?
+* `list` - return a list, which is a special kind of table used for code.
+* `sym` - turn a string into a symbol.
+* `gensym` - generates a unique symbol for use in macros, accepts an optional prefix string.
+* `<symbol>#` - generates a unique symbol, derived from the given symbol.
```
rktjmp, line 1280: @andreyorst `gensym` & `<symbol>#`?

reference.md
andreyorst: on a separate note, while this example shows how macros work, it doesn't show why they're necessary. I mean `plus-10` can be a function without a problem. And this also may encourage people to write macros where a function will do, which is an antipattern.

I think there was an example macro in the tutorial, so perhaps we could instead just link to it here?

rktjmp: Yeah, in OP^

> Macro explanation uses plus-10, which might infer you should be replacing plus-10 functions with macros. Not sure if that needs a note or not.

I agree it is not "a good macro" but I think you want a example where working out what the macro does isn't difficult.

There isn't a an explicit `macro` in the (current?) tutorial on gh but the function could be used:

I think you really want a one liner though.

```
; from tutorial when ctrl-f macro
(fn print-and-add [a b c]
  (print a)
  (+ b c))
```

reference.md
```diff
@@ -1005,6 +1002,39 @@ macros into smaller scopes as well. Note that macros are a
...
+
+(macro print-even [n]
+  (if (= (% n 2) 0)
+    `(print ,n)))
+
+```
+
+As a very simple example, the following macro will emit an expression that adds
+10 to a given argument. Note: this example is *intentionally simplistic* and
+`(plus-10 ...)` would be as function in pratice.
```
rktjmp, line 1022: Explicitly note it's a poor macro?

reference.md
rktjmp: I added a note that it's a poor macro, can't think of a simpler example that doesn't make you think about what the macro is doing while being better served as a function.

andreyorst: > There isn't a an explicit macro in the (current?) tutorial on gh but the function could be used:

Sorry, there's a macro example in https://github.com/bakpakin/Fennel/blob/main/reference.md#import-macros-load-macros-from-a-separate-module, but we were talking with @technomancy on IRC about putting a bit more info to the tutorial

reference.md
andreyorst: I'd say we wait until @technomancy adds some notes on this section

reference.md
```diff
@@ -488,7 +488,7 @@ Example:
 ```
 
 
-Supports destructuring and multiple-value binding.
+Supports destructuring and multiple-value binding, see [`let`](#let-scoped-locals).
```
technomancy, line 491: It makes sense to reference `let` from `local` but you can't use `let` where you would use `var`, so I don't think it's needed here.

reference.md
```diff
@@ -1005,6 +1002,39 @@ macros into smaller scopes as well. Note that macros are a
...
+Macros should return a "quoted" expression (via `` ` ``) to be used as the
+replacement. Normally that expression would "unquote" a value (via `,`) but it
+does not have to. You may not "unquote" a value outside of a "quoted"
+expression.
+
+```fennel
+
+(macro print-even [n]
+  (if (= (% n 2) 0)
+    `(print ,n)))
```
technomancy, line 1016: For this example to work, we would need to explicitly point out that `n` must be a compile-time literal. If you don't understand macros then you would expect `(print-even (+ 1 1))` to work, but it would fail to compile.

reference.md
technomancy: This list is intended to show all the functions which are available in compile scope. While auto-gensym notation is useful and needs to be explained, it should be explained somewhere else.

technomancy:
Thanks for the suggestions. I'd love to expand the macro explanation, but as discussed in chat, I think that belongs in a separate document. Can you remove the macro examples section and the autogensym bit so we can merge the rest of this patch as-is, then we can talk more about how to organize a more thorough guide to macros as its own document?

reference.md
rktjmp: Actually don't think I meant to push this one... :flushed:

rktjmp:
- Removed let links, was intended just to link to de-structuring help, but may be confusing. Perhaps de-structuring could have its own heading in future.
- Removed macro changes

PR #389, opened 2021-09-09
Relax (i)collect :into clause rules

mrichards42:
As discussed in irc:

1. Allows `:into` to appear anywhere (including as the first element in the binding table)
2. Relaxes type validation from "a literal table" to "a literal table, list, or symbol" -- so, excluding numbers or strings
3. Also added an assert that there's only one `:into` clause

### DIFF
--- a/src/fennel/macros.fnl
+++ b/src/fennel/macros.fnl
@@ -115,7 +115,11 @@ encountering an error before propagating it."
         (do (assert (not into) "expected only one :into clause")
             (set into (table.remove iter-tbl (+ i 1)))
             (table.remove iter-tbl i))))
-  (assert (or (not into) (table? into)) "expected table in :into clause")
+  (assert (or (not into)
+              (sym? into)
+              (table? into)
+              (list? into))
+          "expected table, function call, or symbol in :into clause")
   (or into []))
 
 (fn collect* [iter-tbl key-value-expr ...]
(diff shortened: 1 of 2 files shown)


technomancy:
Sorry--I replied a while back but github started silently rejecting my email for some reason.

Thanks for this patch. I've applied the second and third commit.

> 1. Allows `:into` to appear anywhere (including as the first element in the binding table)

I think this might have been a miscommunication. The :into clause is
only documented as being accepted after the iterator; it's an
implementation detail that it's accepted before the iterator. It's coded
the way it is because it's impossible to tell the position of the
iterator until the :into clause has been removed because an iterator can
return any number of values.

I think a better solution would be to step back from the end of the
iterator binding sequence by twos and check for supported clauses
(currently only :into is supported) and remove them from the
sequence. Once supported clauses have been removed, then the last thing in
the iterator binding sequence must be the iterator itself.

technomancy:
Going to go ahead and close this PR. If you would like to make additional changes to make the `:into` clause checks more precise that would be fine but let's do it somewhere else. Thanks!

mrichards42:
> I think this might have been a miscommunication. The :into clause is
> only documented as being accepted after the iterator; it's an
> implementation detail that it's accepted before the iterator. It's coded
> the way it is because it's impossible to tell the position of the
> iterator until the :into clause has been removed because an iterator can
> return any number of values.

That's fair, although I don't think there would be ambiguity in allowing :into to appear before the iterator, since you can't use a string in either the binding position or in the iterator position. But the second commit was the important piece in this PR, so I'm fine with how things stand.

technomancy:
It's not really a question of ambiguity; it's more one of consistency. When you bind values to an iterator, the names of the locals should come right before the thing they're bound to; same as with `each` or `let`. Shoving unrelated elements in between the local and the value disrupts the flow of the code; putting it at the end does not.

PR #393, opened 2021-09-13
warn when a --native-module does not contain a luaopen_* symbol

Benaiah:
Outputs a warning when you specify a `--native-module` that doesn't include a `luaopen_*` function.


technomancy:
Thanks; I think this should be really helpful!

PR #394, opened 2021-09-13
allow native libraries with version extensions

Benaiah:
e.g., `libtinfo.so.6`

### DIFF
--- a/src/fennel/binary.fnl
+++ b/src/fennel/binary.fnl
@@ -235,12 +235,14 @@ int main(int argc, char *argv[]) {
     (os.exit 0)))
 
 (fn native-path? [path]
-  (match (path:match "%.(%a+)$")
-    :a path
-    :o path
-    :so path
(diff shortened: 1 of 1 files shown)


src/fennel/binary.fnl
```diff
@@ -235,12 +235,14 @@ int main(int argc, char *argv[]) {
...
 
 (fn native-path? [path]
-  (match (path:match "%.(%a+)$")
-    :a path
-    :o path
-    :so path
-    :dylib path
-    _ false))
+  (let [(extension version-extension) (path:match "%.(%a+)(%.?%d*)$")]
+    (if (and version-extension (not (version-extension:match "%.%d+"))) false
```
Benaiah, line 239: Actually, I think this needs to check for `version-extension` being an empty string here as well.

technomancy:
Nice catch; thanks.

PR #395, opened 2021-09-13
Allow renaming native modules in compile binary

Benaiah:
Allows using `--rename-native-module ORIGINAL NEW` to rename the require string used to require a given native module from `ORIGINAL` (which is taken from the suffix of the `luaopen_*` symbol(s) in the module) to `NEW`.

This is based on #393 and will be rebased onto `main` once that is merged.

### DIFF
--- a/src/fennel/binary.fnl
+++ b/src/fennel/binary.fnl
@@ -280,7 +280,8 @@ int main(int argc, char *argv[]) {
       (when (= :--rename-native-module (. args i))
         (let [original (table.remove args (+ i 1))
               new (table.remove args (+ i 1))]
-          (tset native.rename-modules original new))))
+          (tset native.rename-modules original new)
+          (table.remove args i))))
     (when (< 0 (length args))
       (print (table.concat args " "))
       (error (.. "Unknown args: " (table.concat args " "))))
--- a/src/fennel/binary.fnl
+++ b/src/fennel/binary.fnl
@@ -179,12 +179,11 @@ int main(int argc, char *argv[]) {
                        "Did you mean to use --native-library instead of --native-module?")
                    :format path)))
         (each [_ open (ipairs opens)]
-          (let [maybe-rename (. rename open)
-                require-name (if maybe-rename
-                                 (do
-                                   (tset used-renames open true)
-                                   maybe-rename)
-                                 open)]
+          (let [require-name (match (. rename open)
+                               renamed (do
+                                         (tset used-renames open true)
+                                         renamed)
+                               _ open)]
             (table.insert out
                           (: "  int luaopen_%s(lua_State *L);" :format open))
             (table.insert out (: "  lua_pushcfunction(L, luaopen_%s);" :format
--- a/src/fennel/binary.fnl
+++ b/src/fennel/binary.fnl
@@ -163,8 +163,11 @@ int main(int argc, char *argv[]) {
     (f:close)
     lua-code))
 
-(fn native-loader [native]
-  (let [nm (or (os.getenv :NM) :nm)
+(fn native-loader [native options]
+  (let [opts (or options {:rename-modules {}})
+        rename (or opts.rename-modules {})
+        used-renames {}
+        nm (or (os.getenv :NM) :nm)
         out ["  /* native libraries */"]]
     (each [_ path (ipairs native)]
       (let [opens []]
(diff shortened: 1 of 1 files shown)


Benaiah:
I'm a little iffy on the help string addition here and the name of the option - there isn't actually a 1:1 correspondence between native modules and the `luaopen_*` symbols within them, so this might be kind of confusing. Any suggestions for better terminology to use for this?

src/fennel/binary.fnl
```diff
@@ -265,15 +276,22 @@ int main(int argc, char *argv[]) {
...
       (print (table.concat args " "))
       (error (.. "Unknown args: " (table.concat args " "))))
     native))
 
 (fn compile [filename executable-name static-lua lua-include-dir options args]
-  (let [{: modules : libraries} (extract-native-args args)]
-    (compile-binary (write-c filename modules options) executable-name
+  (let [{: modules : libraries : rename-modules} (extract-native-args args)
+        opts {: rename-modules}]
+    (each [key val (pairs options)]
```
technomancy, line 293: We have `utils.copy` which would remove the need to copy manually.

src/fennel/binary.fnl
```diff
@@ -304,6 +322,11 @@ To include C libraries that contain Lua modules, add --native-module path/to.so,
 and to include C libraries without modules, use --native-library path/to.so.
 These options are unstable, barely tested, and even more likely to break.
 
+If you need to change the require name that a given native module is referenced
+as, you can use the --rename-native-module ORIGINAL NEW. ORIGINAL should be the
+suffix of the luaopen_* symbol in the native module. NEW should be the string
+you wish to pass to require to require the given native module.
```
technomancy, line 329: I wonder if an example would help here? I know you ran into the need for this in some commonly-used library so maybe clarifying *why* you might need this would be helpful?

src/fennel/binary.fnl
```diff
@@ -176,16 +179,23 @@ int main(int argc, char *argv[]) {
...
-                               open))
-          (table.insert out
-                        (: "  lua_setfield(L, -2, \"%s\");\n"
-                           ;; changing initial underscore breaks luaossl
-                           :format
-                           (.. (open:sub 1 1)
-                               (-> (open:sub 2)
-                                   (: :gsub "_" "."))))))))
+          (let [maybe-rename (. rename open)
+                require-name (if maybe-rename
```
technomancy, line 183: This might be nicer as a match.

technomancy (commented):
Sorry for the delay on this. I think it's a great change but I have a few small suggestions; what do you think?

Benaiah:
I've updated the PR with your review comments in mind - thanks!

What do you think about `--rename-native-symbol` instead of `--rename-native-module`? It's not perfect, because you're not renaming based on the whole symbol (you're renaming based on the suffix of the symbol after `luaopen_`), but it seems to fit better than `--rename-native-module` due to the issue that there isn't a 1:1 correlation between C modules and `luaopen_*` symbols.

technomancy:
> it seems to fit better than --rename-native-module due to the issue that there isn't a 1:1 correlation between C modules and luaopen_* symbols.

TBH I don't have a great understanding of this problem. At first glance it seems to be specifically about changing the name by which the C module is loaded, or in other words, what string is passed to `require`. Can you tell me more about this mismatch? Are you saying a single `.so` file could have multiple `luaopen_*` symbols but only a single Lua module in it?

The other changes look great; thanks.

Benaiah:
> Are you saying a single .so file could have multiple luaopen_* symbols but only a single Lua module in it?

Not quite. One `luaopen_*` function will always correspond to one module/require path, but a single C library can contain more than one module. See the following quote from http://www.lua.org/manual/5.1/manual.html#pdf-package.loaders, particularly the section starting with "The fourth searcher":

>  The third searcher looks for a loader as a C library, using the path given by the variable package.cpath. For instance, if the C path is the string
>
>      "./?.so;./?.dll;/usr/local/?/init.so"
>
> the searcher for module foo will try to open the files ./foo.so, ./foo.dll, and /usr/local/foo/init.so, in that order. Once it finds a C library, this searcher first uses a dynamic link facility to link the application with the library. Then it tries to find a C function inside the library to be used as the loader. The name of this C function is the string "luaopen_" concatenated with a copy of the module name where each dot is replaced by an underscore. Moreover, if the module name has a hyphen, its prefix up to (and including) the first hyphen is removed. For instance, if the module name is a.v1-b.c, the function name will be luaopen_b_c.
>
> The fourth searcher tries an all-in-one loader. It searches the C path for a library for the root name of the given module. For instance, when requiring a.b.c, it will search for a C library for a. If found, it looks into it for an open function for the submodule; in our example, that would be luaopen_a_b_c. With this facility, a package can pack several C submodules into one single library, with each submodule keeping its original open function.

This means that you could pass `--native-module nm.a`, but still have that file contain multiple modules such as `luaopen_nm_a` and `luaopen_nm_b` which you could require as `(let [a (require :nm.a) b (require :nm.b)] ...)`.

(As an aside, I think we should make sure that the behavior of `--compile-binary` matches this behavior as closely as possible. I think we're close, but it's possible that we could figure out a way to make the require paths we output into the binary correspond better with the require paths used to load the native module when it's not statically compiled in. In particular, this would obviate the need for `--rename-native-module` when it comes to the lua readline bindings.)

Given this, I'm kind of torn between `--rename-native-module` and `--rename-native-symbol`. You are indeed renaming a native module, but it doesn't necessarily correspond 1:1 with the files you pass to `--native-module`. Maybe it's best to leave it as is (`--rename-native-module`) and thoroughly document what's going on.

technomancy:
Thank you for the detailed explanation. Let me see if I understand:

* When you say `--native-module` you are specifying a top-level module which can contain submodules inside it under a different `luaopen` symbol.
* When you use `--rename-native-module` you can rename either a top-level module or a submodule.

I think that since the thing being renamed remains a module, it's fine to keep the current terminology you have in this patch. I don't believe it's necessary that because "native-module" means top-level in one context that it must mean top-level in every context.

> As an aside, I think we should make sure that the behavior of --compile-binary matches this behavior as closely as possible. I think we're close, but it's possible that we could figure out a way to make the require paths we output into the binary correspond better with the require paths used to load the native module when it's not statically compiled in.

I'm not sure I follow this but it sounds like it could be a follow-on patch and doesn't need to be resolved in order for this to be merged.

PR #396, opened 2021-09-19
use FENNEL_SRC env variable when building fennel

Benaiah:
This lets you build fennel while outside of the fennel root directory by setting `FENNEL_SRC` to the appropriate path.

Do you know of anywhere else we should be using this if it's provided?

### DIFF
--- a/src/fennel.fnl
+++ b/src/fennel.fnl
@@ -149,9 +149,13 @@
 (set utils.fennel-module mod)
 
 ;; Load the built-in macros from macros.fnl.
-(let [builtin-macros (eval-compiler
-                       (with-open [f (assert (io.open :src/fennel/macros.fnl))]
-                         (.. "[===[" (f:read :*all) "]===]")))
+(let [builtin-macros
+      (eval-compiler
+        (let [FENNEL_SRC (and os os.getenv (= :table (type os)) (= :function (type os.getenv))
+                              (os.getenv :FENNEL_SRC))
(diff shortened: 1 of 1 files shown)


technomancy:
Would `make -C path/to/fennel/checkout fennel.lua` cover what you need here, or is this used outside a context where you can do that?

Benaiah:
`make -C` doesn't fit my needs because I'm not compiling an existing `make` target. Here's the full invocation, if that helps explain what my issue is:

```makefile
fennel-static-http: lua-readline/C-readline.a fennel fennel/fennel
	FENNEL_DEBUG=1 FENNEL_SRC="$$(pwd)/fennel" FENNEL_PATH="$$(pwd)/fennel/src/?.fnl" LUA_PATH="$$(pwd)/lua-readline/?.lua" CC_OPTS="-static -no-pie" \
		./fennel/fennel --lua lua5.3 --no-compiler-sandbox --compile-binary \
		./fennel/src/launcher.fnl fennel-static-readline \
		/usr/lib/lua5.3/liblua.a /usr/include/lua5.3 \
		--native-module lua-readline/C-readline.a \
		--rename-native-module readline C-readline \
		--native-library /usr/lib/libreadline.a \
		--native-library /usr/lib/libhistory.a \
		--native-library /usr/lib/libncursesw.a
```

I can't just `cd` into the `fennel` directory because then my `LUA_PATH` setting wouldn't be correct.

technomancy:
No worries then; cool. Applied the patch.

PR #401, opened 2021-10-20
rationale.md: Fix broken link

elimisteve:


technomancy:
Same as #402 this is a full URL now.

PR #402, opened 2021-10-20
values.md: Fix broken link to Code of Conduct

elimisteve:


technomancy:
I ended up doing a full URL pointing to fennel-lang.org instead for this. I'd be happy to keep it as a local reference if we could get it working in both the local git checkout and the web site, but I'm probably not going to get around to doing that myself.

PR #403, opened 2021-10-27
Destructuring for key value tables

eko234:
This is an implementation for key value table destructuring giving presedence to fennelrest.

### DIFF
--- a/test/core.fnl
+++ b/test/core.fnl
@@ -219,7 +219,7 @@
                "(let [[a & b] (setmetatable {} {:__fennelrest #42})] b)" 42
                "(let [[a & b] (setmetatable {} {:__fennelrest #false})] b)" false
                "(let [[a & b] (setmetatable [1 2 3 4 5] {:__fennelrest (fn [t k] [((or table.unpack _G.unpack) t (+ k 1))])})] b)" [3 4 5]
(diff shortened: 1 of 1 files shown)


technomancy:
Thanks for submitting this.

Conceptually I'm going back and forth on this... It doesn't seem to add *that* much value because `&as` is already available, but on the other hand it is pretty clear what `&` means in the new context by analogy to `&` in sequences. So it's not adding a lot of conceptual footprint to the language; the new burden in terms of understanding is very small, which is good.

I'm curious to see what others say about it.

There are a couple problems though; one is that nowhere else in the compiler do we emit a big string of Lua code like that. Beyond just the style issues, doing so messes up the indentation of the compiled output.

The other is that the commit history of this branch is not clean; if we are to merge this it should just be a single commit branched from upstream main.

eko234:
Oops, yeah the commit history is messy, thanks for reviewing the pr, I do agree that the lua string is not the prettiest of things too.

Thanks again, lets see what people say.

andreyorst:
> There are a couple problems though; one is that nowhere else in the compiler do we emit a big string of Lua code like that. Beyond just the style issues, doing so messes up the indentation of the compiled output.

Do you think we'll ever migrate destructuring to fennel side, as we've discussed on IRC? If so, this will not be an issue, and this code will just be a part of the macro.

PR #404, opened 2021-11-10
Create development container

dbready:
I have become a big fan of [development containers](https://code.visualstudio.com/docs/remote/containers) and created one for Fennel development.

This container installs a copy of Lua 5.1, 5.2, 5.3, 5.4, and JIT. Each version of Lua has an accompanying installation of luarocks.

Each environment lives under `/opt` (`/opt/lua51`, `/opt/lua52`, etc). Each interpreter is symlinked into `/usr/local/bin`:

`/opt/lua51/bin/lua` → `/usr/local/bin/lua5.1`
`/opt/lua52/bin/lua` → `/usr/local/bin/lua5.2`
`/opt/lua53/bin/lua` → `/usr/local/bin/lua5.3`
`/opt/lua54/bin/lua` → `/usr/local/bin/lua5.4`
`/opt/luajit/bin/luajit` → `/usr/local/bin/luajit`

`lua` defaults to pointing to Lua5.4 and `luarocks` points to a Lua5.4 version

The end result of this is that you can boot up the devcontainer and immediately do a `make testall` against all versions, run `make count` utilizing `cloc`, or run `make coverage` which utilizes LuaCov. I was tempted to get `fnlfmt` running inside here, but that seemed a bit too chicken and egg for me.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,53 @@
+# See here for image contents: https://github.com/microsoft/vscode-dev-containers/tree/v0.205.1/containers/debian/.devcontainer/base.Dockerfile
+
+# [Choice] Debian version (use bullseye or stretch on local arm64/Apple Silicon): bullseye, buster, stretch
+ARG VARIANT="buster"
+FROM mcr.microsoft.com/vscode/devcontainers/base:0-${VARIANT}
+
+RUN apt-get update && export DEBIAN_FRONTEND=noninteractive \
+    && apt-get -y install --no-install-recommends \
+    cloc \
+    libc6-dev \
+    libreadline-dev \
+    gcc \
+    make
+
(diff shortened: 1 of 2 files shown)


technomancy:
Thanks for this. I think it's good that everyone be able to install what they need in order to develop.

However, I'm afraid that adding files like this to the Fennel repository itself would give the wrong idea, that the containers are officially supported, when in fact none of the core developers use containers. Perhaps we could link to a separate repo from the setup guide for people who want to develop with containers, and that would make the status clearer.

dbready:
Fair enough. This was a drive-by commit. Also, devcontainers are hardly standardized, and we could see N different styles take off in the next few years.

Happy to see this hosted somewhere else. It took some amount of fiddling to get the different environments to cooperate, which would be nice to have as a reference.

technomancy:
No problem. If you have any suggestions for how to improve the documentation for contributors please let me know.

PR #411, opened 2022-02-07
Detect Lua VM runtime version

cody271:
While attempting to benchmark some Fennel modules against different Lua VMs, I quickly became confused as to what was running where.

Now the Fennel REPL and `--version` CLI flag will perform a best effort detection of the Lua VM environment hosting Fennel.

Also useful API for displaying an "About" dialog in your Fennel app:

```fennel
(fennel.runtime-version)
```

```lua
print(fennel.runtimeVersion())
-- > Fennel 1.0.0 on PUC Lua 5.4
```

Tested against all major PUC versions, latest LuaJIT, Fengari Node CLI, and Fengari Web.


technomancy:
Looks great; thanks! I've applied the patch.

PR #415, opened 2022-03-10
Add embedded fennel documentation

iMarv:
Added a sentence on how to pass compiler options to the embedded fennel compiler.
Using fennel to configure/script the luakit browser and was running into the issue of unknown globals

Finding this method was not easy or quick and is possibly a common issue you can face when plugging fennel into a lua codebase.


technomancy:
Good catch; thanks!

PR #418, opened 2022-03-27
Adding readline support for the REPL public API

gbaptista:
Today, if you start a REPL through the `fennel` command, it will add [readline](https://pjb.com.au/comp/lua/readline.html) support, if available.

But, if you start the REPL through the public API, e.g.:

```bash
fennel -e "(let [f (require :fennel)] (f.repl {}))"
```

Or try to start a REPL inside some part of your code:
```fnl
(local fennel (require :fennel))

(fennel.repl {})
```

The readline support won't be added, even if you have it available.

This PR moves the `try-readline` function from `launcher.fnl` to `repl.fnl` so that both ways add readline support.

If the user provides `readChunk` or `registerCompleter`, It will not add readline support to avoid interfering with the desired customization.

Also, it removes the requirement of the options parameter when starting a REPL: I found no reason to be something mandatory.

### DIFF
--- a/src/fennel/repl.fnl
+++ b/src/fennel/repl.fnl
@@ -294,9 +294,43 @@ For more information about the language, see https://fennel-lang.org/reference")
     (when (not= :exit command-name)
       (loop))))
 
-(fn repl [options]
+(fn try-readline! [opts ok readline]
+  (when ok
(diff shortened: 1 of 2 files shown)


technomancy:
Looks good; thanks!

PR #419, opened 2022-03-28
Adding hashfn missing example

gbaptista:
While upgrading my syntax highlighting package, I missed this example of possible hashfn literal in the reference:

`#{:a $1 :b $2 :c $3}`

```fnl
(local fennel (require :fennel))

(-> [2 4 6]
  (table.unpack)
  (#{:a $1 :b $2 :c $3})
  (fennel.view)
  (print)) ; => {:a 2 :b 4 :c 6}
```


technomancy:
Thanks! I merged this in fcfcb8d but it looks like Github doesn't realize it's the exact same patch.

PR #420, opened 2022-03-28
Exposing varg?, macroPath, and macroSearchers in the Public API

gbaptista:
The function `varg?` is expected to exist in the public API as `fennel.varg?`. The aliases `macroPath` and `macroSearchers` are important for Lua.

I believe that there is an intention to make the public API reliable. Removing or exposing something by accident may lead to breaking things or having to support things that we don't want to because people started to use them.

So, I'm proposing the creation of a kind of _contract testing_ for the public API. wdyt?

Also, some questions:

1) Should every function in the public API be available for Lua as well? If yes, would make sense to create the missing aliases?

```
comment?        => isComment
list?           => isList
macro-path      => macroPath
macro-searchers => macroSearchers
sequence?       => isSequence
sym-char?       => isSymChar
sym?            => isSym
symChar?        => isSymChar
varg?           => isVarg
```

2) May `make_searcher` be removed or deprecated in favor of `makeSearcher`?

Answers from the discussion on IRC:

1) `macroPath` and `macroSearchers` should, the others are unlikely to be used.

2) Too late for that, [make_searcher](https://github.com/search?l=Lua&q=make_searcher&type=Code) is already spread.

### DIFF
--- a/src/fennel.fnl
+++ b/src/fennel.fnl
@@ -105,6 +105,7 @@
             :comment utils.comment
             :comment? utils.comment?
             :varg utils.varg
+            :varg? utils.varg?
             ;; parsing
             :sym-char? parser.sym-char?
             :parser parser.parser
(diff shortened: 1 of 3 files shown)


technomancy:
Thanks for catching this; merged!

PR #423, opened 2022-04-29
Use name-free method for determining parent module

joshbode:
This change is to documentation only, suggesting a method for determining the parent module that doesn't depend upon the current module's name, which should make the technique more reusable.
~Also, for the purpose of demonstration this is split into two steps to make it more obvious what is happening.~


andreyorst:
Hi! The reason for the current approach is that we can't store module name in a variable, as it will break `--require-as-include`, so your suggestion will not work in cases when user want to ship a single lua file script.

joshbode:
ah - I understand. Wouldn't the current solution in the docs that uses `gsub` fail to work for `--require-as-include`, too?

joshbode:
This works as expected, with the code being inlined:
```shell
$ cat libs/example/module-a.fnl
;; file libs/example/module-a.fnl
(local greet (require :libs.example.utils.greet))
(fn hello [] (print "hello from a"))

{:hello hello :greet greet}
$ fennel --require-as-include --complile libs/example/module-a.fnl
local greet
package.preload["libs.example.utils.greet"] = package.preload["libs.example.utils.greet"] or function(...)
  local function greet(who)
    return print(("hello " .. who))
  end
  return greet
end
greet = require("libs.example.utils.greet")
local function hello()
  return print("hello from a")
end
return {hello = hello, greet = greet}
```

However, with the `gsub` expression, as in the docs it doesn't inline the code:
```shell
$ cat libs/example/module-a.fnl
;; file libs/example/module-a.fnl
(local greet (require (: ... :gsub "(.*)%.module%-a$" "%1.utils.greet")))
(fn hello [] (print "hello from a"))

{:hello hello :greet greet}
$ fennel --require-as-include --compile libs/example/module-a.fnl
local greet = require((...):gsub("(.*)%.module%-a$", "%1.utils.greet"))
local function hello()
  return print("hello from a")
end
return {hello = hello, greet = greet}
```

joshbode:
of course, neither does this :)
```
(local greet (require (.. (: ... :match "(.+)%.[^.]+") :.utils.greet)))
```

andreyorst:
This is because you're compiling the wrong file:

```
$ for f in $(find -type f -name '*.fnl'); do sepcat $f; done
= ./libs/example/utils/greet.fnl ==========================================================================
(fn greet [who] (print (.. "hello " who)))
= ./libs/example/module-a.fnl =============================================================================
(local greet (require (: ... :gsub "(.*)%.module%-a$" "%1.utils.greet")))
(fn hello [] (print "hello from a"))

{:hello hello :greet greet}
= ./init.fnl ==============================================================================================
(local a (require :libs.example.module-a))

{:hello-a a.hello}

$ fennel --require-as-include -c init.fnl 
local a
package.preload["libs.example.utils.greet"] = package.preload["libs.example.utils.greet"] or function(...)
  local function greet(who)
    return print(("hello " .. who))
  end
  return greet
end
package.preload["libs.example.module-a"] = package.preload["libs.example.module-a"] or function(...)
  local greet = require("libs.example.utils.greet")
  local function hello()
    return print("hello from a")
  end
  return {hello = hello, greet = greet}
end
a = require("libs.example.module-a")
return {["hello-a"] = a.hello}
```

Your version also works this way.

andreyorst:
We have tests for relative require, can you add your method to them so we can be sure that what we suggest in the docs always works? But please don't remove the old one from the test, this thing is tricky and may have some corner cases, which I can't think of right now. 

I'll also try your method in some of my libraries, which currently use the documented approach and return with the feedback

joshbode:
I've added some tests duplicating the existing `:gsub` method (which is _slightly_ different that what's in the current docs) and didn't have any issues.

joshbode:
I'm not necessarily proposing this, but this macro generalises the concept and appears to work with `--require-as-include`:
```fennel
(macro require* [name]
  (let [n (length (name:match "^%.+"))
        args ["(.+)"]]
    (for [i 1 n] (tset args (+ i 1) "%.[^.]+"))
    (local pattern (table.concat args))
    `(require (.. (: ... :match ,pattern) ,name))))

(local greet (require-rel :.utils.greet))
(local greet (require-rel :..example.utils.greet))
```

andreyorst:
I've tried this approach on my [condition system](https://gitlab.com/andreyorst/fennel-conditions) library and this is a macro I've got to work best:

```fennel
(macro rel-require [module macro?]
  `(.. (or (: (or ... "") :match "(.+%.)[^.]+")
           (if (= ... ,(if macro? "init-macros" "init"))
               ""
               (.. ... ".")))
       ,module))
```

I'll document here why all of this noise is required.

First, I've changed the match pattern a bit from `"(.+)%.[^.]+"` to `"(.+%.)[^.]+"`. Notice that `%.` is now inside the group. This is needed because we never want to start a module name with a dot, as if we require if tom an `init` module, which so happens to be in the same working directory, it will become `(require ".some-module")` which works, but is not guaranteed to work by Lua.

Second, if the module doesn't match it will return `nil` so an `or` is required. This happens when the module doesn't contain a `dot` or when the module is `init`. In case when module doesn't contain a `dot` we simply take it and append a dot to it `(.. ... ".")`, and if it is `init` it means that we're in the same working directory as the required module, so we can use a non-prefixed path, so we return an empty string.

`macro?` is used when we want to do relative require in a module named `init-macros`.

Why would we want to require 'init` directly is a valid question. The answer is, that I (for instance) do it for testing purposes, because I run tests from the library directory, and require the whole library as I test its public API.

Now, still this isn't a general solution, because we have macro modules too. In [cljlib](https://gitlab.com/andreyorst/fennel-cljlib) I use the following relative require trick:

``` fennel
(import-macros {: defn} (if ... (if (= ... :init) :init-macros ...) :init-macros))
```

This is much less involved, as I simply check for `init` (again, for testing purposes) and if it is `init` I return `init-macros` since we're in the same working directory. Otherwise it returns `...` since macros are in `iit-macros.fnl` we can require module directory as is without appending anything. Outer `if` is to make sure that if `...` is `nil` we just use `init-macros`. This is needed because `fennel.dofile` doesn't propagate `...` in compiler scope, which I've reported before.

TL;DR

Macro above works for general cases, but not for macro cases, which needs its own handling.

andreyorst:
altrnatively, a `rel-path` macro can be provided so a user can do `(require (rel-path :foo-bar))` instead, thus making it possible use this with an `import-macros`

technomancy:
Thanks; this looks good, and it's always nice to have better test coverage. Of course as you've noted, it would be much better to have some built-in mechanism instead of copy-pasting this flurry of punctuation around. As Andrey mentioned, we have an open issue for that here: https://todo.sr.ht/~technomancy/fennel/103

The main thing to note here is that the proposed macro should *not* expand to a call to `require`; it should just expand to a module name. That way it's more general and can be used with require or other contexts which need a module name. But we can continue that discussion in the other issue.

Issue #428, opened 2022-06-07
Inconsistent handling of commandline arguments


dokutan:
When running a Fennel script with commandline arguments, some arguments are treated as arguments to the interpreter.

## The problem
Take the the following script (arg.fnl) for example:
```
#!/usr/bin/env fennel
(while (. arg 1)
  (print (. arg 1))
  (table.remove arg 1))
```
When called as ``./arg.fnl -l`` or ``fennel arg.fnl -l``, the following error occurs:
```
runtime error: bad argument #1 to 'open' (string expected, got nil)
stack traceback:
  [C]: in function 'open'
  /usr/bin/fennel:5050: in function ?
  [C]: in function 'xpcall'
  /usr/bin/fennel:5709: in function 'dosafely'
  /usr/bin/fennel:5748: in function 'handle_load'
  /usr/bin/fennel:5801: in main chunk
  [C]: in ?
```
However ``./arg.fnl --version`` or ``fennel arg.fnl --version`` results in:
```
--version
```

## Comparison with Lua
Compare this behaviour to Lua, when arg.fnl is compiled to arg.lua:
```
while arg[1] do
  print(arg[1])
  table.remove(arg, 1)
end
return nil
```
``lua arg.lua -l`` prints ``-l``, ``lua -l arg.lua`` treats ``-l`` as an argument to Lua, resulting in an error.

## Intended behaviour
In my opionion Fennel should only accept arguments that appear before the first filename, all other arguments should be passed to the script. When calling a script directly as an executable no arguments should be handled by Fennel. This would make the behaviour of Fennel identical to Lua.

technomancy:
Good catch; I agree.

Would you be up for submitting a fix?

dokutan:
> Would you be up for submitting a fix?

Sure, i have opened #429.

technomancy:
Fixed in e84c167

PR #429, opened 2022-06-07
Ignore all arguments after the first filename

dokutan:
Fix for #428

### DIFF
--- a/src/launcher.fnl
+++ b/src/launcher.fnl
@@ -90,68 +90,81 @@ If ~/.fennelrc exists, it will be loaded before launching a repl.")
   (match (. arg i)
     :--lua (handle-lua i)))
 
-(for [i (length arg) 1 -1]
-  (match (. arg i)
-    :--no-searcher (do
-                     (set options.no-searcher true)
-                     (table.remove arg i))
-    :--indent (do
-                (set options.indent (table.remove arg (+ i 1)))
-                (when (= options.indent :false)
-                  (set options.indent false))
-                (table.remove arg i))
-    :--add-package-path (let [entry (table.remove arg (+ i 1))]
-                          (set package.path (.. entry ";" package.path))
-                          (table.remove arg i))
-    :--add-fennel-path (let [entry (table.remove arg (+ i 1))]
-                         (set fennel.path (.. entry ";" fennel.path))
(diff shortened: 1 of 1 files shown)


dokutan:
Currently this breaks options that require an argument like ``--add-fennel-path``. I am working on a fix.

dokutan:
06740fab262860ce2a8d84952d1aa912bcb838e2 fixes arguments that were broken by 23b6cfbb17f54f41191fd38367894135f24cafad.

dokutan:
``make lint`` and ``make test`` are passing, there are no infinite loops and my tests described in #428 pass as well. This has turned into a mess.

src/launcher.fnl
```diff
@@ -90,68 +90,85 @@ If ~/.fennelrc exists, it will be loaded before launching a repl.")
...
+        :--eval (set options.i (+ options.i 2))
+        :-e (set options.i (+ options.i 2))
+        :-v (set options.i (+ options.i 2))
+        :--version (set options.i (+ options.i 2))
+        :--help (set options.i (+ options.i 2))
+        :-h (set options.i (+ options.i 2))
+        "-" (set options.i (+ options.i 2))
+        _ (do
+            (set options.i (+ options.i 2))
+            (set options.ignore-options true))))))
```
technomancy, line 171: Thanks for submitting this; sorry for the delay. I broke my hand a few weeks ago.

Instead of every recognized command being a match clause, let's add them all to a `commands` table and then only set ignore-options to true if it's not found in the commands table. It should tidy this up a lot.

Plus instead of adding a new `if` for options.ignore-options, why not merge that into the `while` condition above?

dokutan: > Thanks for submitting this; sorry for the delay. I broke my hand a few weeks ago.


The delay is no problem since i only returned from travelling today. I hope you get well soon.

Thanks for the feedback, i made the changes you suggested in d91bfe3140354a726c459b640d8c6d46aeec9c01. Additionally i used the scope provided by the ``let`` for the ``commands`` table to replace ``options.i`` with a local ``var``.  Please let me know if there is anything else.

technomancy:
Thanks! I've merged and added a test and changelog entry in 29f8d1d.

PR #430, opened 2022-06-18
Pass correct filename to relative require expressions

rktjmp:
When using "relative requires", the module name and module filename are passed into the code chunk when evaluating the "mod-name expression".

Currently this always uses the filename of the first expression, which ends up being `src/fennel/macros.fnl`. We can get the correct filename (the filename for the file that actually contains the "mod-name expression") from the next part of the AST.

It's probably pretty rare that someone wants to operate on the *filename* given inside the expression of a require, but it may as well be correct.

### DIFF
--- a/src/fennel/specials.fnl
+++ b/src/fennel/specials.fnl
@@ -1061,9 +1061,9 @@ Only works in Lua 5.3+ or LuaJIT with the --use-bit-lib flag.")
     (fn find-in-path [start ?tried-paths]
       (match (fullpath:match pattern start)
         path (match (try-path path)
-          filename filename
-          (nil error) (find-in-path (+ start (length path) 1)
-                                    (doto (or ?tried-paths []) (table.insert error))))
+               filename filename
+               (nil error) (find-in-path (+ start (length path) 1)
+                                         (doto (or ?tried-paths []) (table.insert error))))
         _ (values nil
                   ;; Before Lua 5.4 it was necessary to prepend a \n\t to the
                   ;; error message. In newer versions doing so causes an empty
--- a/src/fennel/macros.fnl
+++ b/src/fennel/macros.fnl
@@ -357,8 +357,13 @@ Example:
     ;; this is weird because require-macros is deprecated but it works.
     (let [(binding modname) (select i binding1 module-name1 ...)
           scope (get-scope)
-          macros* (_SPECIALS.require-macros `(import-macros ,modname)
-                                            scope {} binding1)]
+          ;; if the module-name is an expression (and not just a string) we
+          ;; patch our expression to have the correct source filename so
+          ;; require-macros can pass it down when resolving the module-name.
+          expr `(import-macros ,modname)
+          filename (if (list? modname) (. modname 1 :filename) :unknown)
+          _ (tset expr :filename filename)
+          macros* (_SPECIALS.require-macros expr scope {} binding1)]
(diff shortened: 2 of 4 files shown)


src/fennel/specials.fnl
```diff
@@ -1149,12 +1149,21 @@ modules in the compiler environment."
...
+  ;;
+  ;; The first statement in the AST will be an internal function such as
+  ;; `import-macros`, which wont have an appropriate filename to give to the
+  ;; lua chunk.
+  ;;
+  ;; We instead pull the filename from the user-suppied expression. If the user
+  ;; only gave a literal string the filename will be nil, but as that
+  ;; expression is static, that will be inconsequential.
+  (let [{2 second} ast
+        filename (or (and (utils.list? second) second.filename) :unknown)
```
rktjmp, line 1166: Changed from `table?` to `list?` as the argument will always be a list? At least, it never seems to be a plain table.

src/fennel/specials.fnl
```diff
@@ -1149,12 +1149,21 @@ modules in the compiler environment."
...
   ;; and `import-macros`.
-  (let [filename (or filename (and (utils.table? second) second.filename))
+  ;;
+  ;; The first statement in the AST will be an internal function such as
+  ;; `import-macros`, which wont have an appropriate filename to give to the
+  ;; lua chunk.
+  ;;
+  ;; We instead pull the filename from the user-suppied expression. If the user
+  ;; only gave a literal string the filename will be nil, but as that
+  ;; expression is static, that will be inconsequential.
```
rktjmp, line 1164: What I mean here if unclear:

```
import-macros {...} :macros
```

`filename` will be nil, but `:macros` never attempts to use the value so it wont matter.

```
import-macros {...} (.. ... + :x)
```

`filename` will be correct here, and the user could destructure `...` into `modname filename`.

MuhammedZakir: Typo: "user-suppied" -- "suppied" -> "supplied"

technomancy:
This looks sensible, and I think it will get us the desired effect.

However, it seems to be working around a weird implementation detail of `import-macros`. What if instead of working around this quirk, we changed `import-macros` to propagate the filename from the original AST rather than discarding it?

src/fennel/specials.fnl
```diff
@@ -1149,12 +1149,21 @@ modules in the compiler environment."
...
-  (let [filename (or filename (and (utils.table? second) second.filename))
+  ;;
+  ;; The first statement in the AST will be an internal function such as
+  ;; `import-macros`, which wont have an appropriate filename to give to the
+  ;; lua chunk.
+  ;;
+  ;; We instead pull the filename from the user supplied expression. If the
+  ;; user only gave a literal string the filename will be nil, but as that
+  ;; expression is static, that will be inconsequential.
+  (let [{2 second} ast
```
technomancy, line 1165: This style of destructuring was used previously because we were pulling out numeric and string keys from a single table. However, in this case it would be better to use `[_ second] ast` instead since we got rid of the string key.

rktjmp:
> However, it seems to be working around a weird implementation detail of import-macros. What if instead of working around this quirk, we changed import-macros to propagate the filename from the original AST rather than discarding it?

Much smarter, updated.

technomancy:
Thanks; I've merged this!

PR #431, opened 2022-06-23
Added rest destructuring for key value tables

eko234:


technomancy:
Thanks! I've merged this along with a little refactoring.

PR #432, opened 2022-06-23
Make `fennel --eval` print all return values, fix test-arg

emmyposs:
makes dosafely return the same number of values as the function it calls.
fixes test-arg not matching correct output.

this probably?? could use a test for the multiple return from dosafely, id be willing to implement one if yall could give some pointers

### DIFF
--- a/src/launcher.fnl
+++ b/src/launcher.fnl
@@ -52,11 +52,17 @@ If ~/.fennelrc exists, it will be loaded before launching a repl.")
 
 (local options {:plugins []})
 
+;; Lua 5.1 doesn't have table.pack
+;; necessary to preserve nils in luajit
+(fn pack [...]
+  (doto [...]
+    (tset :n (select :# ...))))
+
 (fn dosafely [f ...]
   (let [args [...]]
-    (match (xpcall #(f (unpack args)) fennel.traceback)
-      (true val) val
-      (_ msg) (do
+    (match (pack (xpcall #(f (unpack args)) fennel.traceback))
+      [true &as all] (unpack all 2 all.n)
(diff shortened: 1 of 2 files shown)


technomancy:
Thanks for this patch.

Since `dosafely` is an internal function to the launcher, testing it directly isn't important. As far as I can tell, the only place the return value of `dosafely` is relevant is for `--eval`, right? In that case we can test it in `test/cli.fnl` which is where the other launcher features are tested.

I'm not clear on why the addition of the `pack` function is necessary here. It seems that the same effect could be achieved by just wrapping the `xpcall` in a table, right?

technomancy:
Sorry; I realized my fix for the test conflicts with this so you might want to rebase on main and just include the one launcher commit.

emmyposs:
rebased, added a test for multiple return values.

> [...] It seems that the same effect could be achieved by just wrapping the xpcall in a table, right?

that is mostly correct for PUC lua, but not luajit - luajit stops counting at the first nil regardless of how the table is created, so the number of values needs to be saved.
(note 'mostly correct for PUC lua' - it also stops at the first nil, but only if the table was created from a literal whose last 2 values are nils.)

technomancy:
> that is mostly correct for PUC lua, but not luajit - luajit stops counting at the first nil regardless of how the table is created, so the number of values needs to be saved.

> print(unpack({5, nil, nil, nil, 8, nil, nil}))
> 5

Good lord; that's deeply bizarre and somewhat frustrating. Thanks for the heads up; I had no idea.

I've merged your changes; thanks! (github has a bug where it marks this as closed despite the changes being merged because they were merged outside github)

PR #434, opened 2022-12-05
`case` documentation

rktjmp:


reference.md
```diff
@@ -499,18 +540,124 @@ variables are missing:
...
+    [a b] (+ a x)) ; -> (+ 10 99)
+
+  (case [99 20]
+    ;; [99 20] = [99 b]
+    (where [(= x) b]) (+ x b)) ; -> (+ 99 20)
+
+  ;; PLEASE REVIEW
+  (case [99 20]
+    ;; [99 20] = [99 x]
+    (where [(= x) x]) (+ x x)) ; -> (+ 20 20)
```
rktjmp, line 626: I think this is correct as the user has explicitly opted in to rebind x to something else, if they want to do `(+ 99 20)` then they should have used different symbol name!

reference.md
```diff
@@ -499,18 +540,124 @@ variables are missing:
...
+
+  ;; PLEASE REVIEW
+  (case [10 20]
+    ;; (< (+ 10 20) 99)
+    (where [x y] (< (+ x y) (= x))) (+ x y)) ; -> (+ 10 20)
```
rktjmp, line 631: As above, user as decided to clobber upval.

reference.md
```diff
@@ -499,18 +540,124 @@ variables are missing:
...
+    (where [(= x)]) x
+    _ :no-match)) ; -> 1
+
+(let [pass :hunter2]
+  (case (user-input)
+    (where (= pass)) :login
+    _ :try-again!))
+```
+
+You must pin any symbol you wish to use in a where expression, but not in the
```
rktjmp, line 593: "where" expression?

reference.md
```diff
@@ -499,18 +540,124 @@ variables are missing:
...
   :body)
 
 ;; ok
-(match [1 2 3]
+(case [1 2 3]
   (where (or [a b 2] [a b 3]) (< a 0) (<= b 1))
   :body)
 ```
 
+Symbols bound inside a `cases` pattern are independent from any existing
```
rktjmp, line 555: fixme: `case`

reference.md
```diff
@@ -404,31 +403,90 @@ two distinct concepts with similar names.
...
+
+The final clause will only match if `mytable` has 1 as its first element and
+two non-nil values after it; if so then it will add up the second and third
+elements.
+
+If no clause matches, the form evaluates to nil.
+
+Patterns can be tables, literal values, or symbols. Any symbol is implicitly
+checked to be not `nil`. Symbols can be repeated in an expression to check for
+the same value.
```
rktjmp, line 430: I think this nil behaviour important to outlay upfront  so I moved it up.

andreyorst:
I feel like a more elaborate explanation of the distinction between `case` and `match` is needed, because `case macro that skips unification` doesn't explain much. Though the original `match` explanation wasn't really clear on that part too.

rktjmp:
Yep, haven't quite gotten to `match` yet. Unsure whether to introduce the term `unification` or call it `implicit pinning`, or instead term `pin` as `unify` in `case`...

I don't think `pin` is immediately obvious to a new reader, but I think `unification` might also imply more mechanics. It's more like `references` or `tie` (maybe not `tie`...). 

Ideally there would be one term, `unify operator` `unification` or `pin operator` `pinning` or ... I only call it `pin` because I have seen it in Elixir, not sure of other prior art - that might help get the right (or at least traditional) name.

andreyorst:
A bit off-topic, but, I've been looking at `where`, and in reality, `where` actually acts almost like `and`, with the exception that it only accepts two arguments, the first being the pattern, and the last being the clause.

Maybe we could lift this and allow multiple patterns/clauses and rename it to `and`? As far as I understand we'll get plain `or` in `match` soon, so it will play nicely. @technomancy WDYT?

rktjmp:
> Maybe we could lift this and allow multiple patterns/clauses and rename it to and?

Discarding `or` for a moment, multiple patterns and clauses, or a pattern and multiple clauses? I assume you could only give multiple patterns with `or`.

ex:

A `case` clause accepts multiple patterns in an `or` form. `or` must contain *at least*
one pattern, as described in the preceeding section.

```fennel
(case x
  (or 1 2 3) true ; x was 1 2 or 3
  _ false)
```

```fennel
(case x
  (or 1) :valid
  (or y) :valid
  (or (v-1 v-2) [(= key)]) :valid
  (or) :invalid
  _ false)
```

`case` clauses may also optionally be "guarded" with an `and` form. The first
value/form/expression/argument provided to `and` must be a pattern or
collection of patterns as described by `or` above. Each additional expression
represents a guard. All guard expressions must evaluate to truethy for the
`case` clause to match.

Guard expressions may contain any valid Fennel code and have access to any
in-scope binding.

```fennel
(case x
  (and (or 1 2 3) (sunday? today)) true ; 1 2 or 3 on a sunday
  ;; 1 2 or 3 in a leap year and we rolled lucky
  ;; note that the second `and` is a regular boolean expression
  (and (or 1 2 3) (and (leap-year? today) (feeling-lucky?))) true
  true :ok
  _ false)
```

```fennel
(case x
  (and 1) :valid ;; (and 1 true) but I think worth supporting
                 ;; for easy of modification when writing code.
  (and (or (= x))) :valid
  (and {:count c :name (= user-name)} (and (<= limit c)
                                           (authorised? user-name))) :valid
  (and) :invalid
  (and (or))) :invalid
```

reference.md
```diff
@@ -404,31 +403,90 @@ two distinct concepts with similar names.
...
+  [a ?b] :maybe-one-maybe-two-values
+  ;; maybe-nil == maybe-nil
+  [?a ?a] :maybe-none-maybe-two-same-values
+  ;; maybe-nil, maybe-nil
+  [?a ?b] :maybe-none-maybe-one-maybe-two-values)
+```
+
+Symbols may also be prefixed with `_`, which is similar to `?` except repeated
+symbols values are not compared. `_` is also often used on its own at the end
+of an expression to define an "else" style fall-through.
```
rktjmp, line 464: "repeated symbols values" a bit ugly. I mean that `_x _x` are not compared for equality.

technomancy: Maybe it would be clearer to say that `_foo` means it's completely ignored, rather than highlighting the similarity to `_foo`.

reference.md
```diff
@@ -404,31 +403,90 @@ two distinct concepts with similar names.
...
+Example:
+
+```fennel
+(case mytable
+  ;; not-nil, maybe-nil
+  [a _b] :maybe-one-maybe-two-values
+  ;; maybe-nil, maybe-nil (different to the last example!)
+  [_a _a] :maybe-none-maybe-one-maybe-two-values
+  ;; maybe-nil, maybe-nil, note this is effectively the same as []
+  [_a _b] :maybe-none-maybe-one-maybe-two-values
```
technomancy, line 475: Should clarify here that this clause cannot ever match, because anything that would match here would also match the previous one.

reference.md
```diff
@@ -499,18 +542,154 @@ variables are missing:
...
+
+(let [pass :hunter2]
+  (case (user-input)
+    (= pass) :login
+    _ :try-again!))
+```
+
+Unification is only required inside the pattern. Outer bindings are
+automatically available inside guards and bodies as long as the name has not
+been rebound in the pattern.
```
technomancy, line 600: I'm not sure what "Unification is only required inside the pattern" here means.

I also think we may need to re-evaluate the way we use the term "unification" as it has both a more narrow meaning than how we use it (https://lists.sr.ht/~technomancy/fennel/%3C3j4SZ0Dbj8uiJhME0MPDVUxqBNlNEg1IY9Lv3LPWP_ZZb7EyEgSMvxRXb_TF_TzpT2zGSKs-cIoMuYYMFYrzCcfBE6YAD36i-rvSUuWSU0I%3D%40protonmail.com%3E) and in other cases a broader meaning: https://blog.fogus.me/2010/12/14/unification-versus-pattern-matching-to-the-death/

Perhaps we need to be very specific that we mean "unification with outer scope"; that might be enough.

reference.md
```diff
@@ -148,10 +148,10 @@ Note that the Lua runtime will fill in missing arguments with nil when
 they are not provided by the caller, so an explicit nil argument is no
 different than omitting an argument.
 
-The `lambda` and `match` forms are the only place where the `?foo`
-notation is used by the compiler to indicate that a nil value is
-allowed, but it is a useful notation to communicate intent anywhere a
-new local is introduced.
+The `lambda`, `case`, `match` and `match-try` forms are the only place where
```
technomancy, line 151: The branch now contains `case-try` as well.

reference.md
```diff
@@ -499,18 +542,154 @@ variables are missing:
...
+Unification is only required inside the pattern. Outer bindings are
+automatically available inside guards and bodies as long as the name has not
+been rebound in the pattern.
+
+Example:
+
+```fennel
+(let [limit 10]
+  (case [9 9]
+    ;; automatically avaliable in the guard
```
technomancy, line 607: I'm not sure this is helpful; it seems to just be pointing out that "scoping and variables work like normal here", right? Shouldn't that be the default assumption?

reference.md
```diff
@@ -499,18 +542,154 @@ variables are missing:
...
+    (where [a b] (< limit (+ a b))) :over-the-limit
+    ;; also available in the body
+    [a b] (+ a b limit))) ; -> :over-the-limit
+```
+
+If a pattern defines a binding with the same name as an existing binding in the
+outer scope, the inner binding takes precedence.
```
technomancy, line 614: Let's say instead:

    If a pattern binds a local with the same name as an existing local in the outer scope, the inner binding shadows the outer one.

The "shadows" terminology I think is a little more widespread for this. It might be helpful to reference forward here and mention that this is the difference between `case` and `match`; what do you think?

reference.md
```diff
@@ -499,18 +542,154 @@ variables are missing:
...
+
+Example:
+
+```fennel
+(let [x 99]
+  (case [10 20]
+    ;; x is not unified, so it's a new binding
+    [x y] (+ x y)) ; -> (+ 10 20)
```
technomancy, line 622: I've found that another meaning of "unification" can be internal to a pattern; so a pattern of `[x x]` would also count as unification. Because of this, I think we should prefer "pinned" for this, or possibly "unified with outer scope", but that's kind of a mouthful.

reference.md
```diff
@@ -499,18 +542,154 @@ variables are missing:
...
+    ;; it in a guard or body.
+    (where [p-x y] (< (+ p-x y) x)) (+ p-x y))) ; -> (+ 10 20)
+```
+
+```fennel
+(let [pass :hunter2
+      2fa-key (current-2fa-key)]
+  (case [(input) (2fa)]
+    (where [(= pass) nil]) :login-but-warn-user-they-are-naughty
+    (where [(= pass) 2fa] (valid-2fa? 2fa 2fa-key)) true
```
technomancy, line 649: The `2fa` in the pattern has the same name as the `2fa` function called two lines above; let's rename the function to `get-2fa` to disambiguate.

reference.md
```diff
@@ -499,18 +542,154 @@ variables are missing:
...
+
+### `match` pattern matching
+
+*(since 0.2.0)*
+
+`match` is conceptually equivalent to `case`, except symbols in the patterns are
+always unified with outer-scope symbols if they exist.
+
+It supports all the same syntax as described in `case` except the unification
+(`(= my-symbol)`) expression, as it is always performed.
```
technomancy, line 665: As mentioned above, I think given the way the term "unification" is used elsewhere in my searches, it would probably be clearer to reserve it for "internal" unification within a pattern, and use "pinning" when referring to outer scope.

reference.md
```diff
@@ -499,18 +542,154 @@ variables are missing:
...
+```fennel
+(let [x 95]
+ (match [52 85 95]
+   [b a a] :no ; because a=85 and a=95
+   [x y z] :no ; because x=95 and x=52
+   [a b x] :yes)) ; a and b are fresh values while x=95 and x=95
+```
+
+Eixisting bindings with the value of `nil` do not follow the `?` prefix rules
+when unified.
```
technomancy, line 680: The `?` prefix rule says that locals prefixed with `?` are allowed to match nil values, which is still true if there is an existing binding. Maybe it would be clearer to say that if it is an existing binding, then the `?` prefix is not necessary because the nil-check is bypassed.

technomancy:
Thank you for this thorough patch!

I have some suggestions inline; please let me know what you think. My understanding of how the term unification is used has changed since I did a bit of research, and I now favor the term "pinning" for how we were using "unification" previously.

I have merged the `better-match` branch into main. Could you rebase off that and maybe squash this down into one commit?

Thanks again; this is great work.

technomancy:
@andreyorst what you're saying about `where` vs `and` is really interesting. However, I'd like to keep this thread focused on the docs for the feature as it currently exists; can we start a separate thread about a potential change to `case` and flesh it out there first?

reference.md
rktjmp: Someone new to pattern matching, the term "bindings", "pinning", etc might wonder the rules so I don't think it hurts to reinforce them. It wont hurt those that know but will help those that wonder.

reference.md
rktjmp: https://github.com/bakpakin/Fennel/pull/434/commits/0cd7a2ffb8785e5d4489542b911ec981b958f12a

reference.md
rktjmp: https://github.com/bakpakin/Fennel/pull/434/commits/b4e9c5898d23c9671680025b9be1a4c7cc0e7890 

I have left the `_a _b` there as parity with the `?a ?b` example but noted its non-functional.

reference.md
rktjmp: https://github.com/bakpakin/Fennel/pull/434/commits/bcc389d823f18bbe264e5f2b0778afdf966b13c6

not 100% this is great.

rktjmp:
I've squashed the old commits but left new changes unsquashed just for review, if all ok I will squash them all down.

reference.md
technomancy: I think this is distracting; this section should focus on how `case` works and not reviewing how scoping works in general. It also seems to imply that there might be somewhere else where scoping doesn't work this way, which is a bit misleading.

Other than that, I think this is ready to go; thanks for that!

technomancy: Perhaps if the pinning code example above added another local in the outer `let` which one of the bodies used, this could be demonstrated more subtly in a way that doesn't call extra attention to it?

technomancy:
Yes, please go ahead and squash this down, then I'll merge. Thanks!

technomancy:
Thanks; I've merged this!

PR #439, opened 2023-02-08
Add comparator generator for when all args are idempotent

RobertCochran:
This adds an optimized comparator generator for when all of the arguments are literals or simple (non-multi) symbols (which almost certainly means a simple lexenv lookup). In this version, don't bother wrapping any of the comparison values in an anonymous function since evaluation of the args is idempotent.

This is actually a targeted fix for reducing the memory usage of [char-starter?](https://github.com/bakpakin/Fennel/blob/main/src/fennel/parser.fnl#L61); I'm trying to run Fennel in a low-memory environment (<= 4MB) and when I tried profiling `require("fennel")` with https://github.com/pmusa/lmprof to see where I could find some optimization opportunities, this was one of the indicated hotspots. Creating and immediately throwing away anonymous functions is not *super great* for memory use, and I noticed that this was a compiler enhancement opportunity: recognize when you have idempotent values and elide the function.

I ran the tests and they still pass. I'm 99% confident that this won't break other code; I was pretty simplistic and conservative about what values were considered idempotent, but my understanding might be flawed or incomplete.

### DIFF
--- a/src/fennel/specials.fnl
+++ b/src/fennel/specials.fnl
@@ -952,15 +952,15 @@ Only works in Lua 5.3+ or LuaJIT with the --use-bit-lib flag.")
   (let [comparisons []
         vals []
         chain (string.format " %s " (or chain-op :and))]
-    (for [i 2 (length ast)]
-      (table.insert vals (tostring (. (compiler.compile1 (. ast i) scope parent
-                                                         {:nval 1})
-                                      1))))
-    (for [i 1 (- (length vals) 1)]
-      (table.insert comparisons
-                    (string.format "(%s %s %s)" (. vals i) op
-                                   (. vals (+ i 1)))))
-    (table.concat comparisons chain)))
+    (fcollect [i 2 (length ast) :into vals]
+      (tostring (. (compiler.compile1 (. ast i) scope parent
+                                      {:nval 1})
+                   1)))
+    (fcollect [i 1 (- (length vals) 1) :into comparisons]
+                            (string.format "(%s %s %s)"
+                                           (. vals i) op (. vals (+ i 1))))
+    (table.concat comparisons
+                  chain)))
 
 (fn double-eval-protected-comparator [op chain-op ast scope parent]
   "Compile a multi-arity comparison to a binary Lua comparison."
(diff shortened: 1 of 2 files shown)


src/fennel/utils.fnl
```diff
@@ -128,6 +128,13 @@ Optionally takes a target table to insert the mapped values into."
   (accumulate [max 0 k (pairs tbl)]
     (if (= :number (type k)) (math.max max k) max)))
 
+(fn every [predicate seq]
+  (each [_ item (ipairs seq)]
+        (when (not (predicate item))
+          ;; use escape hatch for early return optimization
+          (lua "return false")))
```
technomancy, line 135: Can you rewrite this as an `accumulate` with an `:until` clause to skip the early return? We really want to avoid the use of the `lua` form.

Also let's name it `every?` instead since it's a predicate.

src/fennel/specials.fnl
```diff
@@ -946,6 +946,22 @@ Only works in Lua 5.3+ or LuaJIT with the --use-bit-lib flag.")
         [rhs] (compiler.compile1 rhs-ast scope parent {:nval 1})]
     (string.format "(%s %s %s)" (tostring lhs) op (tostring rhs))))
 
+(fn idempotent-comparator [op chain-op ast scope parent]
+  "Compile a multi-arity comparison to a binary Lua comparison. Optimized
+  variant for values not at risk of double-eval."
+  (let [comparisons []
+        vals []
+        chain (string.format " %s " (or chain-op :and))]
+    (for [i 2 (length ast)]
```
technomancy, line 955: This would actually be better as `fcollect` instead of `for` but it occurs to me that we haven't backported that macro to the bootstrap compiler yet. Let me see if I can do that, then maybe you can rebase and switch this up?

src/fennel/utils.fnl
```diff
@@ -316,6 +323,14 @@ Returns nil if passed something other than a multi-sym."
 (fn quoted? [symbol]
   symbol.quoted)
 
+(fn idempotent-expr? [x]
+  "Checks if an object is an idempotent expression. Returns the object if it is."
+  (or (= (type x) :string)
+      (= (type x) :integer)
+      (= (type x) :number)
+      (and (sym? x)
+           (not (multi-sym? x)))))
```
technomancy, line 332: This would be really useful to expose in macro scope. (The built-in macros basically implement their own version of this.) We can see about doing that after merging this PR.

src/fennel/specials.fnl
RobertCochran: I noticed you did your end, so I did as suggested. Using `fcollect` in [fe3020f](https://github.com/bakpakin/Fennel/pull/439/commits/fe3020fc9143db10fdfecd9b64386b13c7628c68).

technomancy:
Thanks! After re-running the fuzz testing it cleared up, so it's not related to this change. I've merged it on in!

PR #440, opened 2023-02-09
Reorder `multi-sym?` checks to save memory

RobertCochran:
Rearrange the checks in `multi-sym?` so that we don't do symbol splitting until we're sure the symbol isn't otherwise disqualified from being a multi-sym. This new check ordering saves memory and time compared to the existing one.


technomancy:
This looks great; thank you. I've merged the changes.

PR #454, opened 2023-06-09
Add Makefile target to build Fennel with LuaJIT

Khachig:
This change adds a Makefile target to build a Fennel binary that runs directly on top of LuaJIT. 
This enables very speedy scripts by default and also allows the use of the LuaJIT C FFI within Fennel programs.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -67,17 +67,30 @@ lint: fennel
 
 BIN_LUA_VERSION=5.4.4
 BIN_LUA_DIR ?= $(PWD)/lua-$(BIN_LUA_VERSION)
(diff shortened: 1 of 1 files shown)


technomancy:
Thanks; this looks useful.

I would prefer it if this could be done with the existing rules by just changing the URL and other parameters, but it looks like that's not viable since luajit's build doesn't accept the same `CC_OPTS`.

I can merge this, but can you change it first so that it uses a tag instead of just building whatever's latest on git? Maybe using curl with a tarball of the release so it's more consistent with the existing flow?

PR #465, opened 2023-09-11
Fix bad code generated for names containing backslashes

### LINKED ISSUE #464: Bad code generated in the REPL if the name of a local contains a backslash
When the name of a local contains a backslash, the REPL generates bad code, because the backslash is not escaped when indexing the ``___replLocals___`` table.

Example:
```fennel
>> (local \ 1)
Bad code generated - likely a bug with the compiler:
--- Generated Lua Start ---
local _5c = 1
___replLocals___['\'] = _5c
return nil--- Generated Lua End ---
```

dokutan:
Fixes #464

### DIFF
--- a/src/fennel/repl.fnl
+++ b/src/fennel/repl.fnl
@@ -34,12 +34,12 @@
 
 (fn splice-save-locals [env lua-source scope]
   (let [saves (icollect [name (pairs env.___replLocals___)]
-                (: "local %s = ___replLocals___['%s']"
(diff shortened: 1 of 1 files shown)


technomancy:
Thanks! This works, but it would be better to use `%q` in `string.format` instead of `%s` since we can get Lua to handle this for us.

dokutan:
Thanks, i entirely missed the `%q` specifier while searching for a solution.

technomancy:
Thanks! I've actually merged this already but Github doesn't recognize that fact for some reason so I have to close it manually.

PR #466, opened 2023-09-29
Fix failing tests.

bb010g:
Tested via `nix build ".#fennelCheckAll"`.

This includes a minor compatibility fix for [Faith](<https://git.sr.ht/~technomancy/faith>) that probably needs to be upstreamed.

### DIFF
--- a/default.nix
+++ b/default.nix
@@ -1,0 +1,7 @@
+(import (let
+  lockFile = builtins.fromJSON (builtins.readFile ./flake.lock);
+  info = lockFile.nodes.flake-compat.locked;
+in builtins.fetchTarball {
+  url = "https://api.${info.host or "github.com"}/repos/${info.owner}/${info.repo}/tarball/${info.rev}";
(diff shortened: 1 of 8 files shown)


technomancy:
Thanks. We will keep the packaging in downstream repos instead of in this one, but we can probably take the second commit.

Can you say a little more about this change in the commit message? Under what circumstances is the change to `include` needed?

I don't understand what the compatibility changes for `faith.fnl` are for either; `load` is never mentioned in the file, and the original `math.pow` exists in all supported Lua versions.

technomancy:
Looking at this more closely, I can see that the actual fix is just a 2-line revert of a mistaken change from 5c803e426a5fff59ace6ab7520d14be63cb93c6e. I've committed this fix in its own commit with a descriptive commit message in order to get the tests passing on main again.

If you want to land some of the other changes to the tests (without the downstream packaging changes) we can talk about those separately. Thanks!

Issue #459, opened 2023-07-14
Compatibility with Pluto


dokutan:
Out ouf curiosity I tried running Fennel 1.3.1 on Pluto 0.6.3. [Pluto](https://github.com/PlutoLang/Pluto) is a fork of Lua 5.4 that claims "99.9% compatibility" with Lua:

- Fennel is not compatible with Pluto, because Pluto adds a few [reserved names](https://pluto-lang.org/docs/For%20Integrators/Compatibility#where-are-the-incompatibilities). It might be useful to add it to the incompatible runtimes in the wiki.
- Simply using an editor to replace `new` and `parent` in the source code is enough to make the Fennel script run on Pluto.
- I managed to make most tests succeed using the same renaming approach, note that the `pluto` binary from the releases page is compiled without support for `io.popen`, building Pluto from source removes that issue.
- When running Fennel on Pluto a few warnings about excessive arguments are produced. I am unable to decide whether these findings should be considered bugs in Fennel.

```
./pluto fennel-1.3.1
fennel-1.3.1:2053: warning: too many arguments [excessive-arguments]
    2053 | provided = safe_compiler_env(false)
         | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ here: expected 0 arguments, got 1.
fennel-1.3.1:2287: warning: too many arguments [excessive-arguments]
    2287 | return add_macros(macro_loaded[modname], ast, scope, parent_)
         | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ here: expected 3 arguments, got 4.
fennel-1.3.1:2414: warning: too many arguments [excessive-arguments]
    2414 | return add_macros(macro_tbl, ast, scope, parent_)
         | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ here: expected 3 arguments, got 4.
fennel-1.3.1:4119: warning: too many arguments [excessive-arguments]
    4119 | parse_string(b)
         | ^^^^^^^^^^^^^^^ here: expected 0 arguments, got 1.
fennel-1.3.1:4987: warning: too many arguments [excessive-arguments]
    4987 | local _135_0, _136_0 = f0(k, x)
         | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ here: expected 1 argument, got 2.
[...]

PR #476, opened 2024-04-02
Add option for additional Lua keywords

dokutan:
This PR adds the `--keyword` option to improve support for Lua forks which have additional keywords, as suggested in #459.

This option only affects the generated code, i.e. fennel generates correct code if this option is used.
Therefore this option will not allow Fennel to run on a Lua fork, if any added reserved names are used by Fennel itself (for example the Pluto keyword `new` is used by Fennel). However it should be possible to compile Fennel with this option to prevent this issue. 

I have tested this with [Ravi](https://github.com/dibyendumajumdar/ravi) and [Pluto](https://pluto-lang.org/):

- Ravi (`--keywords defer,C__decl,C__unsafe,C__new`):
  - Fennel can run on Ravi without changes and generates correct Ravi code
- Pluto (`--keywords switch,continue,enum,new,class,parent,export,try,catch`):
  - Fennel can't run on Pluto without changes but generates correct Pluto code

If this functionality gets added, it leads to a few questions that i can not answer:
- Does it make sense to document the forks with their keywords in the wiki?
- Should this option be added to the bootstrap compiler?
- Should Fennel detect these forks to:
  - show the correct vm name/version in the welcome message?
  - add a notice to use this option in the welcome message?
  - automatically add the keywords?

### DIFF
--- a/src/fennel/utils.fnl
+++ b/src/fennel/utils.fnl
@@ -384,6 +384,9 @@ When f returns a truthy value, recursively walks the children."
                      :while true
                      :goto true})
 
+(fn add-lua-keyword [str]
+  (tset lua-keywords str true))
+
 (fn valid-lua-identifier? [str]
   (and (str:match "^[%a_][%w_]*$") (not (. lua-keywords str))))
 
@@ -480,6 +483,7 @@ handlers will be skipped."
  : string?
  : idempotent-expr?
  : valid-lua-identifier?
+ : add-lua-keyword
  : lua-keywords
  : hook
  : hook-opts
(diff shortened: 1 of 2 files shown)


dokutan:
> However, in order to be consistent with other compiler options, it should be controlled by the options table rather than modifying the list of keywords in utils.fnl.

 I have opened #477 which implements the same functionality using the options table and documented the changes in `api.md` and `man/man1/fennel.1`.

> Without that, it would make it so fennel.lua itself couldn't be loaded into Ravi/Pluto, right?

Ravi works without this by chance, because no added keywords are used in the fennel codebase, however for Pluto this would be necessary.

PR #477, opened 2024-04-02
Add option for additional Lua keywords (using the options table)

dokutan:
Identical to #476, but implemented using the options table.

This change allows correct code to be generated when running Fennel code, compiling Fennel code to Lua and in the repl, including the `,compile` repl command.

### DIFF
--- a/api.md
+++ b/api.md
@@ -39,6 +39,8 @@ usually accept these fields:
 * `unfriendly`: disable friendly compiler/parser error messages.
 * `plugins`: list of compiler [plugins](#plugins).
 * `error-pinpoint`: a list of two strings indicating what to wrap compile errors in
+* `keywords`: a table of the form `{:keyword1 true :keyword2 true}` containing
+  symbols that should be treated as reserved Lua keywords.
 
 You can pass the string `"_COMPILER"` as the value for `env`; it will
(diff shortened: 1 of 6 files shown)


technomancy:
Thanks; this looks great.

I have some minor hesitations about adding `keywords` to the `run-command-loop` function. I think if we do this, the arglist for that function will just grow out of control over time. Unfortunately it is part of the public API due to the fact that plugins can contain repl commands, so we can't remove this later if we want to streamline it.

What if instead of passing `keywords` we pass the whole `opts` table to `run-command-loop` and to the individual command functions? That would leave us more flexibility going forward.

It also might be nice to add a test to `test/mangling.fnl` if you don't mind.

I've merged all the commits except for the repl one into main.

technomancy:
Thinking about this more, it seems that the problem of propagating options from the top-level `repl` call down into the `,compile` command is more general, and the fact that opts are being dropped was already a bug even before the keyword option was added; this just made that problem more obvious.

I'll close this PR out since the main functionality has been applied. Thanks!

technomancy:
I went ahead and fixed the repl bug in 3383463 and added tests in e039b52 so I think we're all good here; thanks again!

Issue #462, opened 2023-08-20
`,apropos-doc` cannot find functions defined in REPL


HiPhish:
Hello,

the `,apropos-doc` and `,apropos-show-docs` cannot find functions defined on the REPL. Here is what I am doing in a fresh Fennel REPL started from the command-line:

```
$ fennel
Welcome to Fennel 1.3.1 on PUC Lua 5.3!
Use ,help to see available commands.
>> (fn herp [] "Herp derp lol" :herp)
#<function: 0x559fe514bcf0>
>> (fn derp [] "Herp derp rofl" :derp)
#<function: 0x559fe4f8e450>
>> ,apropos-doc derp

>> ,apropos-show-docs derp
```

First I define two functions `herp` and `derp` which have similar (but not same) docstrings. Then I search for the pattern `derp`. The first comma-command prints an empty line, the second one does nothing. However, the `,apropos` command can find them just fine:

```
>> ,apropos erp
___repl___.env.___replLocals___.derp	___repl___.env.___replLocals___.herp
```

technomancy:
Yeah, currently those commands only search loaded modules, but they could certainly be extended to also search repl locals. Would be happy to take a patch for this.

PR #481, opened 2024-06-14
Fix bug in apropos REPL commands

wrance:
Fix bug described in https://github.com/bakpakin/Fennel/issues/462 by stripping `_.G` from the input rather than the output of the `apropos*` function. This appears to have been the original intended behavior as described by the comment on [line 206](https://github.com/bakpakin/Fennel/blob/main/src/fennel/repl.fnl#L206).

### DIFF
--- a/src/fennel/repl.fnl
+++ b/src/fennel/repl.fnl
@@ -206,9 +206,7 @@ For more information about the language, see https://fennel-lang.org/reference")
   ;; _G. part is stripped from patterns to provide more stable output.
   ;; The order we traverse package.loaded is arbitrary, so we may see
   ;; top level functions either as is or under the _G module.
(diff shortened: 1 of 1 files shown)


technomancy:
Looks great; thanks! I've applied your patch using git instead of using github's flow, so it doesn't register as merged, but it is there.

I think I will leave #462 open because while this does allow apropos to find the function in question, it shows under `_G.___replLocals___`, which is a little counter-intuitive. But it's a big improvement over before where we didn't show it at all!

Thanks again.

wrance:
Awesome thanks! Is there another place that would be more intuitive to find it under? Or are you just suggesting that that part be stripped from the output string?

technomancy:
Yeah, I just meant the prefix should be excluded since anyone in the repl would be able to access those values by just typing them directly in.

PR #494, opened 2025-03-08
Fix typo in rationale.md.

cqfd:
Just a couple small typos I noticed while reading the docs.


technomancy:
Thanks. I've applied your patch in d790c33 but github did not detect it as applied, so I'm closing manually.

PR #495, opened 2025-03-16
update broken links in lua-primer

crmsnbleyd:
luasocket and luafilesystem both now reside at https://lunarmodules.github.io


technomancy:
Good catch; thank you!

I've merged this in ac75caf but github failed to detect it, so I'm closing manually.

PR #499, opened 2025-04-18
check if arugment to `sym` is a string

### LINKED ISSUE #497: `sym` of a symbol should fail with a better error message, or not fail at all.
Hi, I ran into an unpleasant error message while building a macro like this:

```
(macro ->cell [[& deps] & exprs]
  `(make-computed-cell
     (fn []
       (let ,(accumulate [binds []
                          _ dep (ipairs deps)]
               (do
                 (table.insert binds (sym dep))
                 (table.insert binds `((. ,(sym dep) :get)))
                 binds))
         ,(unpack exprs)))))
```

Esentially I want to turn this:

```
(->cell [a b c] (+ a b c))
```

into something like this:

```
(make-computed-cell
  (fn []
    (let [a (a.get)
          b (b.get)
          c (c.get)]
      (+ a b c))))
```

But that's beside the point.

The issue is that I have a bug on these lines in the macro:

```
  (table.insert binds (sym dep))
  (table.insert binds `((. ,(sym dep) :get)))
```

And when I run the code it fails with a tricky error:

```
>> (->cell [a b c] (+ a b c))

runtime error: attempt to concatenate a table value (local 'raw')
stack traceback:
	...l-1.5.3-1/fennel-1.5.3-1-rocks/fennel/1.5.3-1/bin/fennel:3635: in function 'getname'
	...l-1.5.3-1/fennel-1.5.3-1-rocks/fennel/1.5.3-1/bin/fennel:3717: in function 'destructure_sym'
	...l-1.5.3-1/fennel-1.5.3-1-rocks/fennel/1.5.3-1/bin/fennel:3887: in function 'destructure1'
	...l-1.5.3-1/fennel-1.5.3-1-rocks/fennel/1.5.3-1/bin/fennel:3900: in function 'destructure'
	...l-1.5.3-1/fennel-1.5.3-1-rocks/fennel/1.5.3-1/bin/fennel:1530: in function 'special'
	...l-1.5.3-1/fennel-1.5.3-1-rocks/fennel/1.5.3-1/bin/fennel:3414: in function ?
	...l-1.5.3-1/fennel-1.5.3-1-rocks/fennel/1.5.3-1/bin/fennel:1284: in function ?
	...l-1.5.3-1/fennel-1.5.3-1-rocks/fennel/1.5.3-1/bin/fennel:3414: in function ?
	...l-1.5.3-1/fennel-1.5.3-1-rocks/fennel/1.5.3-1/bin/fennel:3458: in function ?
	...l-1.5.3-1/fennel-1.5.3-1-rocks/fennel/1.5.3-1/bin/fennel:3935: in function ?
	...l-1.5.3-1/fennel-1.5.3-1-rocks/fennel/1.5.3-1/bin/fennel:6083: in function ?
	[C]: in function 'xpcall'
	...l-1.5.3-1/fennel-1.5.3-1-rocks/fennel/1.5.3-1/bin/fennel:6990: in function ?
	[C]: in ?
```

This is doubly confusing because `macrodebug` of the expression gives what looks like the correct result:

```
>> (macrodebug (->cell [a b c] (+ a b c)))

(make-computed-cell (fn {} (let [a ((. a "get")) b ((. b "get")) c ((. c "get"))] (+ a b c))))
```

The problem of course is that `dep` used in the `accumulate` actually already is a symbol and calling `sym` on it will cause the error later on.
`macrodebug` somehow doesn't trigger it.

So I would suggest there are a couple of possible solutions:
1. Add an assertion to the `sym` definition, so that it fails with a better error message when the user tries to make a symbol out of a symbol
2. Detect if a symbol was passed to `sym` and just return it right away

https://github.com/bakpakin/Fennel/blob/adc5b31a3aec9f401f1b24b7559a1811c572c92f/src/fennel/utils.fnl#L203

It seems like it would be pretty easy to do either of them and I'd be happy to contribute a pull request, I just wanted to ask first which solution would be preferred.

msladecek:
Added an assertion to check if the first argument to `sym` is a string.

resolves #497

### DIFF
--- a/src/fennel/utils.fnl
+++ b/src/fennel/utils.fnl
@@ -205,6 +205,7 @@ they are specifically not cons cells."
 not exposed outside the compiler. Second optional argument is a table describing
 where the symbol came from; should be a table with filename, line, bytestart,
 and byteend fields."
+  (assert (= (type str) :string) (.. "sym expects a string as the first argument, received " (type str)))
   (setmetatable (collect [k v (pairs (or ?source [])) :into [str]]
                   (if (= (type k) :string) (values k v)))
                 symbol-mt))
--- a/test/macro.fnl
+++ b/test/macro.fnl
@@ -807,6 +807,20 @@
   (t.= {1 :a 3 :c 5 :e :n 5} (macro-wrap pack (pack :a nil :c nil :e))
        "pack is in compiler-env"))
 
+(fn test-sym []
+  (macro use-sym [arg]
+    `(do ,(sym arg)))
+
+  (t.= "(do something)"
+       (macrodebug (use-sym "something") true)
+       "constructing a symbol from string should not fail")
+
+  (let [(ok? err) (pcall fennel.eval "(do
+                                       (macro use-sym [arg] `(do ,(sym arg)))
+                                       (use-sym something))")]
+    (t.is (not ok?))
+    (t.match ".*sym expects a string as the first argument.*" err)))
+
 {:teardown #(each [k (pairs fennel.repl)]
               (tset fennel.repl k nil))
  : test-arrows
@@ -832,4 +846,5 @@
  : test-case-try
  : test-lambda
  : test-literal
- : test-env-lua-helpers}
+ : test-env-lua-helpers
+ : test-sym}


technomancy:
Applied and pushed; thanks!

